PR #23, opened 2015-02-23
REST Api client lib

mrpollo:
- It supports only GET calls at the moment
- Some of the functions don't match the url endpoints because of how
  python manages variable names (see test-cloud.py for examples on how
to run some calls)
- If a new call has to be added and it includes dots in name add to
  REST_CALLS, and see source for reference

**TODO**:
- [ ] POST Calls syntax
- [ ] Sessions
- [ ] LOGIN
- [ ] SIGNUP
- [ ] Setup codeship to run this new testsuite


mrpollo:
cc @geeksville finally found the time to start working on this.

Also thanks @gussy

mrpollo:
Created a test runner shell script, once we merge we need to add it to codeship, it will run all files that match **"test"** and end on **".py"**
```
./test-runner.sh
```

geeksville:
@mrpollo looks great - can you fix that mds key thing?  and then I'll merge it into master.

mrpollo:
@geeksville i changed things a bit, can you review?

mrpollo:
Also make sure you set an environment variable called DRONEAPI_KEY

geeksville:
looks great - thanks!

PR #29, opened 2015-03-30
fix pip package name

mrpollo:
fix install command on docs, ```droneapi-python``` is not the correct pip package name


geeksville:
thanks!

Issue #36, opened 2015-03-31
Rename "Python DroneKit Doc"


hamishwillee:
This title is meaningless, and could apply to any of our docs in the kit :-)

This would be much better as "DroneKit Python API Reference". OK?

mrpollo:
Ok, thanks!

hamishwillee:
Cool, done in linked merge request #38. Closing this item.

PR #38, opened 2015-03-31
Rename API Reference doc

hamishwillee:
Rename API Reference doc as discussed in #36
Improve the index.rst python docs entry point as discussed in #40 
Fixed pip dependency information (to python-pip) as discussed in #41


mrpollo:
:+1: thanks @hamishwillee

PR #45, opened 2015-04-03
Vagrant for cross-platform doc builds

### LINKED ISSUE #43: Provide information and vagrant file on how to contribute to docs?
We should make it easier for people to get started with contributing to the documents. The good news is that *sphinx_3dr_theme* can now be updated from pip, so it is now possible to set up sphinx and build the source without having to modify any files (previously you needed to modify the conf file in order to replace the theme). 

I've written a vagrant script that does this. The proposal is to put this in /docs/ alongside the .rst files. The (windows) workflow would be:

1. Create a fork of this project
1. Clone the fork locally 
1. Navigate to your docs folder and run "vagrant up". This will bring up the VM, add sphinx and the theme, pull the latest code git, then build (call "make html").
1. Users can then modify the files in their windows file system/notepadd++ and cause a new build by running: ```vagrant ssh  -c "cd /vagrant && make html"```
1. They can commit the changes and do other git activities through windows. (or through the VM).

Is this approach desirable? Is the location of the vagrantfile in docs OK? (I think it is, but just checking).

In addition, I propose to update README.md to add information about how to contribute (specifically to the docs, but more generally it would be useful to cover how people can join the project?).

I would do this by adding a new heading "Contributing" near the end with a link "Contributing to the Documentation". That doc would explain how to use the vagrant file etc. 

OK?

mrpollo:
@hamishwillee brought up the great idea of adding vagrant so any Windows user can help build our docs as well. #43

I left the Vagrantfile on project root mainly because we need to install the requirements.txt pip packages to be able to build the autodocs, this means your 1 liner needs to change to this:

```bash
vagrant ssh -c "cd /vagrant && make html"
```

@hamishwillee please add to this branch the contributing changes on the README, there are also lots of WARNINGS on the doc builds im not sure how to fix those, can you please help?

fixes #43


hamishwillee:
The file itself looks good. 

IFF there might ever be a need for a different set of vagrant tools, it would be better to move this from the root to /tools/doc/

I'll look at the readme changes and build warnings on Tuesday.

Thanks!

mrpollo:
the reason behind it being in root is because we need to install the pip requirements before we can generate the autodocs, also i think having many vagrant images is overkill, what do you think?, going forward we should try to incorporate everything into this vagrant image

hamishwillee:
Thanks for explanation. 

I am completely supportive of the idea of a single vagrant file for "everything" if that is possible - provided we con't compromise on the ease-of-use of the system for different purposes. Would this be used for dronekit image VM too?

W.r.t. the pip requirements I don't really understand - maybe the vagrantfile could take relative file requirements inputs, and indeed you can include requirements in the vagrantfile itself. Given that we'll have only one vagrant file it is not really an issue, and I will assume you've looked into the limitations and know what will work. Thanks!  

Will try look further into warnings and test this on Tuesday.

hamishwillee:
Hi Ramon

I have added "	Tidy heading levels and fix doc build warnings caused by errors in " above. 

There were three sources of DOCS warnings 
- in getting_started, the heading underlines were shorter than the headings. 
- in __init__.py there were some errors in the text definitions read by sphinx (e.g. :returns instead of :return:)
- Images are being imported by URL rather than by file system path

The change fixes the first two warnings. Is there any reason that files like https://github.com/diydrones/droneapi-python/raw/master/example/documentation/drone-delivery-splash.png cannot/should not be in the docs themselves?

The other changes were tidy up to the headings for consistency with our android docs.

There are lots of other warnings, but these are related to the code itself, and not the docs. 

With respect to your above comments, to build the docs we need to do:

    vagrant ssh -c "cd /vagrant/docs && make html"

Otherwise this is great. Would be good if this change went in master. I haven't created docs on how it is used yet. My thought was that I might do it in my own fork and then contribute back to master. This is less confusing to me than messing with the vagrant branch ....

hamishwillee:
PS Is that everything you needed from me?

mrpollo:
@hamishwillee i rebased with master and fixed some conflicts, can you double check everything is in order before merge?

hamishwillee:
@mrpollo Looks good to me - thanks for rebasing (I think this might have been avoided if I had been more careful). Please do commit!

mrpollo:
perfect!! :star: :star: :star:

PR #46, opened 2015-04-04
(Docs) Moved supported companion computers to bigger link

phmagic:


mrpollo:
Thanks :+1:

hamishwillee:
Agree that the companion computer option needed to be highlighted, and given that we don't really have any information in the docs, pointing off to wiki probably a good idea at this point. 

My only problem is it feels like in the docset we've given up on helping people working with companion computers. I think I'd have been happier having the existing material that you moved turned into own doc "Getting started with Companion Computers" with more and better links into the wiki that we'd eventually build out. Fine for now.

This doesn't feel quite right to me as a solution - we talk about companion computers but we don't in the docset really provide an

Short term this is a good solution - we don't have detail docs in this docset for companion computers so linking elsewhere makes sense. I would have slightly more happy having

PR #58, opened 2015-04-09
Fixes the API Reference and also tidies some layout issues in the contributing docs.

hamishwillee:


hamishwillee:
Hi @jmacgithu, the commit https://github.com/hamishwillee/dronekit-python/commit/89f693e305b775c1bb3f400ed00785e1473a88fe tidies up the text of your example (#69). I think it is improved, but feel free to comment/suggest modifications.

@mrpollo , when do you plan on merging my commits - we're over 21 days behind? Probably I should create separate branches, but when I started this you were merging on a daily basis so it seemed unnecessary.

mrpollo:
@hamishwillee can you in the future send PR's for separate issues?, it is really hard to review a PR that has many different changes in it, even if they are all from the same topic, a PR should be: easy to review,  up to date with master and not be under active development after submission.

Issue #62, opened 2015-04-13
test-upload.py is not a dronekit-python example


hamishwillee:
[test-upload.py](https://github.com/diydrones/dronekit-python/blob/master/example/log_upload/test-upload.py) shows how to upload a file to a service with the Cloud (REST) API, using Python. IMO that makes it something that belongs in the cloud example code, NOT in the dronekit-python example code.

When you've created a place for the cloud docs and issues, I'll move this across (assuming you agree)

mrpollo:
lets remove this example

hamishwillee:
I will delete

hamishwillee:
Closed by #82

PR #82, opened 2015-05-06
Tidy Flight Replay example. Remove another invalid example.

hamishwillee:
Delete invalid example as per #62 
Fix flight replay code to use command code

@mrpollo Sorry these are in same pull request - were supposed to be in separate ones. No idea how to split them.


mrpollo:
you probably forked from the branch where you fixed the flight replay no worries I already merged that so it will be ok

:+1:

Issue #86, opened 2015-05-06
GPSInfo   - what values are returned if we can’t get a fix


hamishwillee:
GPSInfo 

- what values are returned if we can’t get a fix?
- Are all values as in https://pixhawk.ethz.ch/mavlink/#GPS_RAW_INT supported (other docs I see don't have RTK or DGPS options)

The documentation in API Reference currently ignores these points.

mrpollo:
if you have no lock you get ```None``` back

hamishwillee:
Fantastic. Thank you.

hamishwillee:
FYI I updated the ``Vehicle.gps_0`` doc to note that this returns None if there is no lock (#90)

PR #90, opened 2015-05-07
Fixes APIRef around GPSInfo and send_mavlink

### LINKED ISSUE #85: message_factory and send_mavlink(message) docs have incorrect advice
``message_factory`` and ``send_mavlink(message)`` state that custom commands are a catch-all/back door in case we miss something in the API. 

This is incorrect. These are also useful for sending arbitrary commands in guided mode. 

----

In addition, we need guide documentation on these. The documentation should cover that there are a number of ways to send messages. For example COMMAND LONG is a basic message format that can be used to send any seven parameter command. So you could use:
```
vehicle.message_factory.command_long_encode to send a MAV_CMD_CONDITION_YAW 
```

However you could also send this command using its own method like below. The benefit being that you don't have to think about command ids, sequence numbers etc.:

```
vehicle.message_factory.mav_cmd_condition_yaw_encode
```

hamishwillee:
Fixes #85 and part of #86.

### DIFF
--- a/droneapi/lib/__init__.py
+++ b/droneapi/lib/__init__.py
@@ -403,7 +403,7 @@ class Vehicle(HasObservers):
 
     .. py:attribute:: gps_0
 
-        GPS position information (:py:class:`GPSInfo`).
+        GPS position information (:py:class:`GPSInfo`). ``None`` if there is no GPS lock.
 
 
     .. py:attribute:: battery_0_soc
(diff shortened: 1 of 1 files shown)


mrpollo:
@hamishwillee you get ```None``` on the attributes of ```gps_0``` not on the actual response from ```gps_0```

```python
vehicle = get_vehicles()[0]
print vehicle.gps_0 # returns GPSInfo class even if no lock available
```
**output**:  ```GPSInfo:fix=None,num_sat=None```

When I said it returned ```None``` I meant on the ```GPSInfo``` attributes

```python
print vehicle.gps_0.eph # None
print vehicle.gps_0.epv # None
print vehicle.gps_0.fix_type # None
print vehicle.gps_0.satellites_visible # None
```
**output** on all of those print statements is ```None```

Can you update your docs on this branch to reflect this please?,

hamishwillee:
@mrpollo Thanks for clarification. The commit I just added fixes this.

mrpollo:
:+1:

PR #94, opened 2015-05-11
Create Getting Started/QuickStart and restructure

hamishwillee:
This request creates the QuickStart and updates the Getting Started. It also completes the restructure of the doc set so that Getting Started and Companion Computers are in the "Guide". This is the final structure of the doc set - from now on we extend the guide and examples appropriately.


hamishwillee:
@mrpollo The last commit https://github.com/hamishwillee/dronekit-python/commit/689ef9bc798766be19a2c049c5b8baf14f1f7454 adds back in the instructions to disable the arming check and load some parameters. We removed this originally as "not necessary" but I now know that the default vehicle configuration is quite unforgiving, and may never arm properly. Our choice if we want to run examples that will arm is then either to disable the checks in the simulation or in the scripts. Much better to do in the simulation so we don't promote dangerous coding practices.

mrpollo:
:+1:

Issue #93, opened 2015-05-09
Move/tidy up /example/documentation


hamishwillee:
The example documentation is currently in two places:

* https://github.com/diydrones/dronekit-python/tree/master/example/documentation
* https://github.com/diydrones/dronekit-python/tree/master/docs/examples

Action is to move the documentation all into the official docs. First step to simply remove the content of migrated docs and add links to where the content is now. When all examples are migrated, then remove the stubs and perhaps just have a readme in root of /master/example/ explaining where the docs are.

@mrpollo - make sense?

mrpollo:
Agreed.

hamishwillee:
This is closed by #96. Note that there are undocumented examples still (e.g. client_sketch.py). I will raise separate issues for those.

PR #96, opened 2015-05-11
Move all example documentation to official documentation set

hamishwillee:
This addresses #93. 

All the example documentation in [/example/documentation](https://github.com/diydrones/dronekit-python/tree/master/example/documentation) has now moved into the official documentation along with images (which were previously linked from docs by URL). The README was updated to point to the official docs as well. 

So after this pull, all docs and doc-related material should be present and maintained in the /docs/ tree.


mrpollo:
btw when creating an issue and a PR to fix that issue the practice is to not close the issue until the PR has been merged, also the PR can auto-close the issue with a PR message or a commit message per https://help.github.com/articles/closing-issues-via-commit-messages/

mrpollo:
:+1: great change!

hamishwillee:
Thanks very much - and for the advice on PRs. I'll follow that now.

Issue #103, opened 2015-05-12
Vehicle.mount_status needs to be documented


hamishwillee:
The mount_status is defined in MPVehicle. If run on my SITL environment the following code returns 

```
print "mount_status: %s" % v.mount_status  #Defined in MPVehicle

mount_status: [None, None, None]
```

This needs to be documented. @mrpollo is best way would be to move this into the public API? Alternatively, I can create documentation in the init__.py file. 

In docs, it would be good to be able to show how to configure SITL with a camera mount so we can get meaningful information back from this.

hamishwillee:
Closed by #108.

Issue #101, opened 2015-05-12
Vehicle.battery_0_volt and Vehicle.battery_0_soc should be hidden in documentation


hamishwillee:
``Vehicle.battery_0_volt`` and ``Vehicle.battery_0_soc`` were in the original documentation but are not implemented. 

These should be hidden in the generated documentation (currently they are just marked as "not implemented") - action for @hamishwillee.

hamishwillee:
Closed by #108 - just deleted them from the docs as they were manually added (not actually attributes in the class definition)

Issue #100, opened 2015-05-12
Remove Vehicle.waypoint_home documentation


hamishwillee:
``Vehicle.waypoint_home`` was in the original documentation, but is not present in the implementation (MPVehicle). So if you try the following you will get an exception.

```
print "waypoint_home: %s" % v.waypoint_home
```
Remove the variable from the documentation as it is "not implemented". 

#105 has been created to add the attribute back again

hamishwillee:
Research indicates that implementing this API this probably isn't possible currently. It is possible to [set the home position in a mission](http://planner.ardupilot.com/wiki/common-planning-a-mission-with-waypoints-and-events/common-mavlink-mission-command-messages-mav_cmd/#mav_cmd_do_set_home) but it is not 100% clear whether this is the same "variable" as is used in other modes. It is also not obvious that there is any way to get the home position. There was a brief discussion on this in the [discussion group some years ago](https://groups.google.com/forum/#!topic/mavlink/SL0oTUJx1fE).

geeksville:
definitely possible, it was probably inadvertently lost when full command access was added..  vehicle.commands[0] is guaranteed to be home.  It would be easy to add it back.

hamishwillee:
@geeksville Thanks very much. 

@mrpollo - do you want to add it back or should we just document vehicle.commands[0] gets the right value? My preference is to add it back to make it more visible. 

While we're at it, can/should we make this have a setter as well? That makes sense to me because then you could easily change the home position outside of missions.

hamishwillee:
Closed by #108

maruina:
I have a couple of doubts:  
* What should be the type of waypoint_home? A Location instance? An array with lat, lon, alt?
* Let's say that I set a waypoint_home and then I create a mission: should my home overwrite vehicle.commands[0] or vice versa? Or should be they always kept separate from each other? Should my waypoint_home be None unless I set a mission?

hamishwillee:
@maruina All good questions. I would say

> What should be the type of waypoint_home? A Location instance? An array with lat, lon, alt?

Yes, a Location instance.

> Let's say that I set a waypoint_home and then I create a mission: should my home overwrite vehicle.commands[0] or vice versa? Or should be they always kept separate from each other? Should my waypoint_home be None unless I set a mission?

[...]

Issue #99, opened 2015-05-12
Error in Vehicle.armed markup


hamishwillee:
[Vehicle.armed](http://python.dronekit.io/automodule.html#droneapi.lib.Vehicle.armed) renders incorrectly. Probably missing a backtick or two.

> his attribute can be used to get and set the armed` state of the vehicle (``boolean).

hamishwillee:
Closed by #108

PR #108, opened 2015-05-13
Fixes to API Reference to remove non-implemented functions from docs

hamishwillee:
This primarily removes attributes and other entities that were not implemented from the documentation (those that were added to documentation, not automatically pulled from the interface definition). It also removes documentation about parameter change notification, which is not supported.

It addresses #101, #100, #103,#89, #99 (see those posts for more information).

I also split down a number of the longer lines to make them a little shorter (cosmetic/layout). This was to improve the readability of the source in github.

### DIFF
--- a/docs/about/overview.rst
+++ b/docs/about/overview.rst
@@ -31,7 +31,7 @@ The API provides classes and methods to:
 
 - Get a list of connected vehicles.
 - Get and set vehicle state/telemetry and parameter information.
-- Receive asynchronous notification of state/parameter changes.
+- Receive asynchronous notification of state changes.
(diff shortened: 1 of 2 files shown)


mrpollo:
:+1: great changes as well

hamishwillee:
@mrpollo Note, this is not live yet - if it should have been rebuilt, there may be a problem with the build.

mrpollo:
there seems to be a problem that I need to fix when we make changes to the docstrings, everything on this pr is now up do date on http://python.dronekit.io/ thanks for the report @hamishwillee

PR #117, opened 2015-05-14
Adds a debugging topic. 

### LINKED ISSUE #104: Add guide topic on debugging
Covering print statements, pdb (Python debugger), anything else we can come up with.

hamishwillee:
This request fixes #104.

### DIFF
--- a/docs/about/contributing.rst
+++ b/docs/about/contributing.rst
@@ -1,3 +1,5 @@
+.. _contributing:
+
 ============
 Contributing
 ============
--- a/docs/guide/debugging.rst
+++ b/docs/guide/debugging.rst
@@ -1,0 +1,91 @@
+.. _debugging:
+
+=========
+Debugging  
+=========
+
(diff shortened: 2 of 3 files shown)


mrpollo:
can you please add a NOTE at the end of the pdb chapter that says something in the lines of "for more information on pdb check the docs on the pdb site".

Can we also add a NOTE in general that says better debugging is an action item that is being worked on and can we create an issue and link to that issue on the this debugging srt on the docs, if people have comments on debugging they can leave feedback there

hamishwillee:
@mrpollo - good suggestion re having issue to track debugging improvements. I added note in the introduction of page for this, linking to #118 

I don't particularly agree with the first note because the first link pointed to the site, and further down I pointed to the debugging commands - adding an explicit note that repeats this link is irritating for readers. Despite my concerns, I pulled the link to the debugger commands and the site down to the end of the text (before the last note). I think that probably addresses your concern. 

Ready for you to pull!

mrpollo:
:+1: thanks!

PR #124, opened 2015-05-18
Guide and examples for working with states and parameters.

### LINKED ISSUE #109: Rename small_demo.py example to vehicle_state.py and document
The [small_demo.py](https://github.com/diydrones/dronekit-python/blob/master/example/small_demo/small_demo.py) example contains information about getting and setting vehicle state and parameters (including the rc_overrides). This should be renamed to explain what the content is actually about. I am proposing "vehicle_state.py" because the name "state" can be used to imply this is about both the dynamic attributes (position, speed etc) and also the parameters (settings - including RC overrides).

The file should have the following updates:
* Enable the mode callback which is currently disabled. Add an "armed" callback. No need to implement other callbacks because we can't change them in this "static" example. Show how to remove both of these.
* Enable a set_mavlink_callback() (currently disabled). Show how to disable this.
* Show all the attributes being fetched (currently does most, but add mount and any other missing attributes)
* Show how to set parameters (currently commented out). Add links to other parameters? Show what happens if unsupported parameter used?
* Remove the "Test custom commands" commented out material
* Add clear section markers/comments to each section.

Add a document about this example in the docset. 

Add a guide document about setting parameters etc.

hamishwillee:
This resolves #109 

In overview, it renames small_demo.py to vehicle_state.py and extends it to cover (almost) all aspects of working with states/parameters. It adds a detailed guide topic describing how the vehicle states/params are used and a "thin" example page too. The quickstart/getting started also updated for the new example name.


hamishwillee:
@mrpollo once this guide is in we need to decide how much information we keep in API reference and how much we keep in guide. We should certainly attempt to reduce the guide material in the reference if this is duplicated.

mrpollo:
can you rebase this please?

hamishwillee:
@mrpollo I think this is correctly rebased now. Good to go.

mrpollo:
:+1:

Issue #97, opened 2015-05-11
Client Sketch example is undocumented


hamishwillee:
The [sketch](https://github.com/diydrones/dronekit-python/blob/master/example/sketch/client_sketch.py) example is undocumented. Evaluate whether this should be documented, and if not, whether it should be removed.

mrpollo:
we should probably remove it

hamishwillee:
The sketch is now deleted, so this can be closed

PR #126, opened 2015-05-18
Delete the client sketch

hamishwillee:
As per  #97, delete the client_sketch.py because it does not run and includes primarily APIs that aren't implemented.


mrpollo:
:+1:

PR #127, opened 2015-05-19
Recommend best modes to use with DroneKit

### LINKED ISSUE #119: Update mode documentation to recommend GUIDED/AUTO
Update [VehicleMode](http://python.dronekit.io/automodule.html#droneapi.lib.VehicleMode) reference to *recommend* AUTO and GUIDED modes. 

At the moment we don't really comment on this other than to point to the supported modes. In practice though you really shouldn't be using dronekit within a manual mode (which may *require* a transmitter) unless you know what you're doing. These modes are supported on all platforms and are known to be safe. 

@mrpollo Seem reasonable?

hamishwillee:
This improves the VehicleMode documentation and  resolves  #119. The interesting text is:

> The flight mode determines the behaviour of the vehicle, and what commands it can obey. DroneKit apps should typically use the AUTO mode for waypoint missions and GUIDED mode otherwise. Care should be taken when using manual modes as these may require remote control input from the user (RETURN_TO_LAUNCH is generally safe!)

This links to the state/parameters guide introduced in #124 (not yet in master), but will build and read OK even if that commit is not yet present.

### DIFF
--- a/droneapi/lib/__init__.py
+++ b/droneapi/lib/__init__.py
@@ -173,14 +173,18 @@ class VehicleMode(object):
     """
     This object is used to get and set the current "flight mode". 
 	
-    The flight mode determines the behaviour of the vehicle, and what commands it can obey. For example, 
-    the AUTO mode is used (on all vehicle platforms) when executing stored waypoint missions.
+    The flight mode determines the behaviour of the vehicle, and what commands it can obey. DroneKit apps should typically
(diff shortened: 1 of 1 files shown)


hamishwillee:
I just made a further update. The instructions for mode use on plane/rover are quite different from Copter both because the vehicles are different and because the implementation of the message handling is different.

mrpollo:
excellent work @hamishwillee

hamishwillee:
The API Ref hasn't built with this change - see #124

PR #134, opened 2015-05-22
Improve copter simple_goto and adds guide and reference for takeoff

hamishwillee:
The previous simple_goto example had a number of problems, including using channel_override for thrust rather than the takeoff command, lack of pre-arm checks. The update does it the right way. I updated the example documentation along with this.

In addition, I created a guide topic for taking off (copter only) and added apireference for the takeoff function (note, this is just in "text" and I still think we need to expose the method in the API properly).

### DIFF
--- a/docs/examples/simple_goto.rst
+++ b/docs/examples/simple_goto.rst
@@ -1,42 +1,133 @@
-========================
-Example: Simple Go To
-========================
+.. _example_simple_goto:
 
-This little demonstration just tells the vehicle to fly to a couple of different locations in the world.  You can edit the code to pick a latitude and longitude close to your position.
+==============================
+Example: Simple Go To (Copter)
+==============================
+
+This example demonstrates how to arm and launch a Copter in GUIDED mode, travel to a number of waypoints, and then return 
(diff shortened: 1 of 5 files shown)


mrpollo:
hey @hamishwillee didn't we remove the examples under /example on a previous PR?, aside from that I am about to merge (but will wait until you reply)

mrpollo:
also you might need to rebase with the latest changes from master again, sorry for this :)

hamishwillee:
@mrpollo No worries, all done. 

>didn't we remove the examples under /example on a previous PR?

No, we removed the legacy example docs under /example/documentation. In previous PR we also changed small_demo.py to vehicle_state.py.

mrpollo:
ok cool, what do you think of renaming ```example``` to ```examples```?, it's one of those things that make me cringe when I see it, maybe it's just my OCD

hamishwillee:
> ok cool, what do you think of renaming example to examples?

I must have OCD too. Only concern is that there are a lot of mentions of this now. I'll sort it out though.

PR #135, opened 2015-05-28
Add main battery support

maruina:
This adds the support for the main battery.
I created a Battery object with voltage, current and level attributes and you can create an observer to update the values.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 build
 .vagrant
 *~
+.idea
--- a/droneapi/lib/__init__.py
+++ b/droneapi/lib/__init__.py
@@ -169,6 +169,26 @@ class GPSInfo(object):
         return "GPSInfo:fix=%s,num_sat=%s" % (self.fix_type, self.satellites_visible)
 
 
+class Battery(object):
+    """
+    Battery info available
+
+    :param battery_voltage: Battery voltage in millivolts
+    :param battery_current: Battery current, in 10 * milliamperes, -1: autopilot does not measure the current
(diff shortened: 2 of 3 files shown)


mrpollo:
@maruina thanks for your great contribution :+1: this will go in the next release

mrpollo:
@hamishwillee I believe the codeblock should be sufficient for docs but you might want to take a look at it

hamishwillee:
@mrpollo - I can't tell - http://python.dronekit.io/automodule.html has reverted to a very old version (again).
Probably there will be more work to be done in the guide - I've created #137 to track that.

mrpollo:
I just updated the docs so you can see them :)

hamishwillee:
@mrpollo thanks very much. I'll need to doc the member in Vehicle class (missing). #137 has been updated with actions. I'll try to get to that today or Monday.

@maruina Nicely done. Thank you.

hamishwillee:
@mrpollo @maruina there is one slight inconsistency in the implementation. The Battery class returns -1 value when  current measuring (for example) is not implemented. The rest of the API tends to return None for items that "don't work".  @mrpollo - up to you to decide whether this really is an inconsistency, and if so, whether/priority on fixing.

maruina:
Thank you, it's a great project.
@hamishwillee I can fix the code returning None if you prefer that way, no problem.

hamishwillee:
@maruina @mrpollo I would prefer None return for consistency. However @mrpollo is the architect, so final say on him!

hamishwillee:
@maruina , please do change this to return None rather than -1. Also please see bug I reported at the end of #137 . When you've updated your material, I'll rebase and update my patch.

PR #147, opened 2015-06-02
Added tip for dronekit install

djnugent:
Just a tip for installing dronekit.

### DIFF
--- a/docs/guide/getting_started.rst
+++ b/docs/guide/getting_started.rst
@@ -34,6 +34,12 @@ Installing DroneKit
 
 *DroneKit* can be installed on Linux, Windows and Mac OSX. 
 
+.. tip::  *pip* will be used to install pacakages. It can be installed by executing the following
+    
+.. code:: bash
+
+    sudo easy_install pip
+
 
 Installing DroneKit on Linux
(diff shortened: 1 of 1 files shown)


hamishwillee:
The following section installs pip as just another component: 
```
sudo apt-get install python-pip python-numpy python-opencv python-serial python-pyparsing python-wxgtk2.8
```
To my mind adding this tip rather than just installing "all dependencies in one" makes things more complicated. 

Can you explain the benefit. Otherwise recommend reject.

djnugent:
Ah I did not see that. This was more in regards to Mac osx. In my experience I had to install pip before I could do anything. It just depends on how much we want to hold the user's hand through the installation.

hamishwillee:
If it is MacOS relevant then it belongs in the OSX section, not up the top. I haven't tested the existing OSX docs. Again though, with docs we try to keep it simple - so if the step isn't needed, lets not have it.

djnugent:
Alright sounds good

hamishwillee:
Thanks for posting it though - always good to reconsider why we've done things a certain way (often the situation changes too!)

PR #149, opened 2015-06-03
Update contributions section with how to set up a dev environment

hamishwillee:
This splits "Contributing" out into its own section and splits up the original docs appropriately. It adds  instructions on how to set up dev environment using Linux/Vagrant.  This is appropriate because we are now getting more interest and contributions, and we need to make it easier for people to contribute.

All the instructions are significantly updated:

* I reduced Daniel's Linux instructions to reference the existing wiki document (which gets updated by the ArduPilot team). This also fixes some limitations with the original doc like the fact that it is Copter-specific.
* I added a very brief style-guide to the documentation section.

I think this is good, but I will need to re-read it tomorrow. At the moment I am too close to the text and would not see defects.


djnugent:
@hamishwillee I don't know how you replied to a specific part of my comment, but I exclusively use "-w" and have no problems. Looking at sim_vehicle.sh it appears "-w" pulls from the same location. https://github.com/diydrones/ardupilot/blob/master/Tools/autotest/sim_vehicle.sh#L367

hamishwillee:
@djnugent You're right - excellent. I mostly use Ramon's vagrant startup script and that ignores the -w command. I'll fix this up.

djnugent:
@hamishwillee I think the reason Ramon doesn't use "-w" is because he starts SITL using a different script than sim_vehicle.sh. However, both do the same thing!

hamishwillee:
@djnugent In theory yes. But in testing shows that the -w option doesn't work using his script. I haven't attempted to debug but I added an issue https://github.com/mrpollo/sitl-vagrant-recipe/issues/1

hamishwillee:
@djnugent As we briefly discussed last night, I had an epiphany -  the only difference between developing *using* DK vs developing *for* DK is that when you're modifying DK itself you need to rebuild your own fork of DK. Thats a very small delta against the instructions in the Quick Start and Getting Started.

So I've cut out all the stuff in the contributions on setting up SITL etc and simply recommended first following Quick Start/Getting started, and then explained how to fork and build DK on top of that for each of the platforms. It is much simpler and more modular. 

Can you please check you're OK with this?

djnugent:
Exactly! That is the only difference. It looks much better! I like the titling better too. "Setting up DroneKit-Python for Development on Linux" was kind of confusing. *That was the question I had about wordage last night*

hamishwillee:
@djnugent Thanks Daniel (and yes, the naming was "clunky"). @mrpollo - probably good to take this in now.

mrpollo:
@hamishwillee @djnugent thanks guys!

PR #156, opened 2015-06-10
MPVehicle: add system_status property

mrpollo:
exposes the ```system_status``` field from the most recently received heartbeat message

handy for checking whether they vehicle is on the ground, flying, in failsafe, etc

possible values:

Field name | Description
---------- | ------------
MAV_STATE_UNINIT | Uninitialized system, state is unknown.
MAV_STATE_BOOT | System is booting up.
MAV_STATE_CALIBRATING | System is calibrating and not flight-ready.
MAV_STATE_STANDBY | System is grounded and on standby. It can be launched any time.
MAV_STATE_ACTIVE | System is active and might be already airborne. Motors are engaged.
MAV_STATE_CRITICAL | System is in a non-normal flight mode. It can however still navigate.
MAV_STATE_EMERGENCY | System is in a non-normal flight mode. It lost control over parts or over the whole airframe. It is in mayday and going down.
MAV_STATE_POWEROFF | System just initialized its power-down sequence, will shut down now.

@hamishwillee can you update this branch with docs to match this change please?

big thanks to @liamstask for the contribution.


mrpollo:
@djnugent can you please review?

droneapi/module/api.py
```diff
@@ -450,6 +455,7 @@ def mavlink_packet(self, m):
             self.level = m.battery_remaining
             self.__on_change('battery')
         elif typ == "HEARTBEAT":
+            self.system_status = m.system_status
             self.__on_change('mode', 'armed')
```
djnugent: Shouldn't 'system_status' be added to the self._on_change()?

mrpollo: I think you are right, let me ponder on this for a bit.

djnugent: I think its only use is to have the ability to add a attribute_observer. I think you should add it to keep consistent will the other attributes.

billbonney:
The thing to remember about mode is that it's quite broad. APM indicates more specifically what is happening in status using the status bits and the custom_mode ie. LAND or RTL indicates probably a failsafe in action.

hamishwillee:
@billbonney Can you point me point me to the status bits documentation? Perhaps we can provide additional states through the API for these. Whatever we do we need to be able to access this information through DroneKit (at the moment since we only have send_message without any way to get/parse a result I don't see how we could get this information).

billbonney:
On looking I only see status bit as part of the dataflash logs. I was under the impression that the onboard_control_sensors_health bits contained this information i.e. when it's landed. I'll talk with Randy/Tridge if there is some message they are sending.

That said, most states are inferred through information from the telemetry. Send Message is like giving instructions to your kids, you are never sure they have heard you, or going to do what you ask, until you see them doing it. ;)

hamishwillee:
@billbonney Thanks very much - interested to find out what you find out. It may be that we can do some of the "inferring" behind the scenes/within the API.

hamishwillee:
@mrpollo What happens if we lose the heartbeat - will this return NONE? 
Is this just a string that gets returned?
I am doing some testing now, but would be good to have canonical answers.

hamishwillee:
@mrpollo OK, as you can see above, documentation added. This includes updates to the example code, example code doc, guide on setting states. I've tried to make it clear that this is broad information by adding (in reference) that "This attribute provides **high-level** information about whether the vehicle is on the ground, flying, in failsafe, etc. ". Depending on what @billbonney comes back with we might add a whole separate guide topic on "How to deal with fail safe conditions"

The vehicle_state.py example code adds a function to print out a human-readable string for the different MAV_STATE values. I think this is enough to show how you would use the API.

Ready to pull.

mrpollo:
Inspired by the conversation I made some small changes to the value of ```system_status```, this attribute now returns a Class ```SystemStatus``` which prints the status description and compares to ints for when we want to do a check against another status.

@hamishwillee would you be able to check the docs I'm not sure that I did it right

hamishwillee:
@mrpollo I updated the docs and then attempted to rebase to master. 

Unfortunately I'm not sure I did this right. The [file comparison](https://github.com/diydrones/dronekit-python/pull/156/files) - as though a lot of new content was discovered. However the files themselves look fine. Can you please sanity check?

tcr3dr:
It appears that the line endings on a few items was incorrect. I fixed them and pushed.

The target of this PR will be the "major" branch once it's updated. I'll take a look sometime next week.

tcr3dr:
Replaced by #441.

PR #157, opened 2015-06-11
Adds wxpython to the requirements for running wx on mac.

tcr3dr:


hamishwillee:
@tcr3dr Awesome, I couldn't test MAC instructions because I don't have one. I *think* you can run without these, but you don't get the console (could be wrong, and in any case, this is good to have installed)

mrpollo:
is the reason behind this for proper wx support in MAVProxy?

hamishwillee:
@mrpollo if that was to me - I don't know. I do know that the Windows instructions are missing these, which is why we still can't run console and map from the dronekit prompt (but we can still run DroneKit commands through the command prompt).

mrpollo:
this was for @tcr3dr, but I think you answered already.

@djnugent can you look into the WX requirements for windows and send a PR for them to the guides?

djnugent:
@mrpollo I will give it a look

tcr3dr:
@hamishwillee We *can* run without those, and we could stipulate that, but the errors when they are not working are obtuse enough (submodule import issues get converted to module missing issues by mavproxy) and brew is so easy. I'd be in favor of requiring / including these steps for the moment.

mrpollo:
this makes sense, thanks for clarification @tcr3dr 

:+1:

hamishwillee:
@tcr3dr Ditto. Couldn't agree more. 

@djnugent would love to see instructions for "all the bits" windows as requested by Ramon above. I understand that to do this you need to do all sorts of horrible things like installing/downloading opencv and then doing a pip install on top of the downloaded content. Rather you than me :-)

djnugent:
@mrpollo @hamishwillee can you clarify what you'd like? A solution on how to display map and console or a guide for installing wx? Installing wx on windows did not help display map and console. Mavproxy just wont locate console and map(at least on my machine). I am unaware of what the "all the bits" are...
capture

hamishwillee:
@djnugent The former. 1. Instructions to be able to install DroneKit on windows and get both the map and console (wx is just one dependency). 

BUT This is going to be a bitch to work out, and then a pain for anyone else to follow. What would be much better is to solve this issue: https://github.com/Dronecode/MAVProxy/issues/148 (and now we loop back to why sorting 148 is useful ! )

djnugent:
Ok  I'm starting to understand the issue. There are two ways to install mavproxy(exe installer and pip install). Each have downsides.

hamishwillee:
@djnugent Exactly! Now if windows could install the dependencies using pip in the same way as we do on Linux then I wouldn't be suggesting including DK in MAVProxy installer. There is some useful information on setting up [MAVProxy on windows here](http://tridge.github.io/MAVProxy/mavdevenv.html).

Issue #159, opened 2015-06-12
Best way to exit a script if you get into trouble?


hamishwillee:
If my battery goes low, or if a failsafe occurs, then it is likely the vehicle will RTL. In this case my script may well still be running. What is the best method to call to exit it/how can I exit it to MAVProxy without killing mavproxy?

djnugent:
```
while not api.exit and vehicle.mode.name != 'RTL:
      <your code here>
```
Your script ends when there is nothing else to execute. Just put everything in a while loop and provide a exit condition.

To kill the script from mavproxy run `api stop <thread #>`

djnugent:
Or for an immediate less clean method you could do this:

```
import sys

<your code before>

if vehicle.mode.name == 'RTL':
     sys.exit()

<your code that wont run after the script is killed>
```

hamishwillee:
@djnugent Thank you very much! The real example I'm looking at is that if my code is running and we go low battery there is an automatic RTL. However my goto code will actually reset the mode to guided and there can be some fighting over control in this case.

The first example isn't really all that useful because the loop is only checked when you enter the block. If you're doing lots of things in the loop then it simply isn't responsive enough. Of course you can use this around every operation, but then you end up having these loops all over the place. Yuk. 

The second looks more useful. I can add a vehicle mode callback and shut the script down from that. I'll try it :-)

As an aside ... we probably need a document on "good coding practices for drone scripts". If we were to write something like that, any other things you would mention? First thing that springs to mind is basic safety - don't do horizontal movement if your vehicle is at person height.

djnugent:
Yeah it really depends on how your script is structured. 
+1 on the good coding practices. 
Explain why `while not api.exit:` is important. 
Explain how to check for command completion for various commands. I love "simple goto" with the arm_takeoff function. That is the proper way to use dronekit and the docs don't explain that right now.

hamishwillee:
> That is the proper way to use dronekit and the docs don't explain that right now.

Working on it :-)  Coding practices would be a good place for this information.

IMO ``while not api.exit:`` is another case where you'd probably be better off having a callback ... if it were observable ... because the API can in theory exit at any time. Do you think we can make it observable?

djnugent:
@hamishwillee I'm sure we could make it observable. I'm not sure how "observe" works(a separate thread??) but it would be nice if the user didn't have to worry about it.

hamishwillee:
[...]

PR #166, opened 2015-06-18
Added api.stop() to support threadicide

djnugent:
Allow the user to cleanly exit the thread from the current script. This partially solves issue #159 

See example usage here https://gist.github.com/djnugent/d65e6adb24344dcdf970

### DIFF
--- a/droneapi/module/api.py
+++ b/droneapi/module/api.py
@@ -294,6 +294,9 @@ class MPAPIConnection(APIConnection):
     def get_vehicles(self, query=None):
         return [ self.__vehicle ]
 
+    def stop(self):
+        raise SystemExit('script commanded itself to stop')
+
 class APIThread(threading.Thread):
     def __init__(self, module, fn, description):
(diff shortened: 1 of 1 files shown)


mrpollo:
Is there any sort of event triggered when the thread dies on its own?

djnugent:
@mrpollo  Nothing special. After a thread ends/dies/suicides it goes here. 
https://github.com/diydrones/dronekit-python/blob/master/droneapi/module/api.py#L336

Just a small clean up for keeping track of threads

hamishwillee:
@djnugent Did you add any API reference for this? 

I think until we've got a way to call this "anywhere" (as per #159) that is probably the extent of what we should do. OK?

tcr3dr:
DK 2.0 won't need this. Sorry @djnugent!

djnugent:
But it had such a catchy name :( I will endure.

hamishwillee:
:-)

PR #167, opened 2015-06-18
Fixed simple_goto.py to exit properly

djnugent:
This prevents the script from continuing to run even after mavproxy has flagged to the the script to stop

### DIFF
--- a/examples/simple_goto/simple_goto.py
+++ b/examples/simple_goto/simple_goto.py
@@ -27,14 +27,16 @@ def arm_and_takeoff(aTargetAltitude):
     while vehicle.gps_0.fix_type < 2:
         print "Waiting for GPS...:", vehicle.gps_0.fix_type
         time.sleep(1)
-		
+        
     print "Arming motors"
     # Copter should arm in GUIDED mode
     vehicle.mode    = VehicleMode("GUIDED")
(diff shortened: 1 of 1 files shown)


examples/simple_goto/simple_goto.py
```diff
@@ -27,14 +27,16 @@ def arm_and_takeoff(aTargetAltitude):
...
     print "Arming motors"
     # Copter should arm in GUIDED mode
     vehicle.mode    = VehicleMode("GUIDED")
     vehicle.armed   = True
     vehicle.flush()
 
-    while not vehicle.armed and not api.exit:
+    while not vehicle.armed:
+        if api.exit: #check to make sure mavproxy hasn't commanded us to stop
+            api.stop()
```
hamishwillee: @djnugent This changes is fine by me. When we have a callback on api.exit things will be much cleaner of course since you won't have to have this checker code everywhere if you're always going to be killing the thread.

tcr3dr:
Now merged into master, the proper way to exit would be closing the connection and raising SystemExit.

hamishwillee:
@tcr3dr 

>the proper way to exit would be closing the connection and raising SystemExit.

Really - example? Should this be done automatically when/if a script completes/falls through?

Issue #114, opened 2015-05-14
DroneKit has no method for detecting command failure - do we need one?


hamishwillee:
Currently our instructions suggest that you change the mode like this:
```
v.mode = VehicleMode("AUTO")
v.flush() # After this, write to vehicle has been made
```
The problem is that after this code the write has succeeded, but the mode may not actually have been changed (and if you read back ``v.mode.name`` it may well be the old name). A better coding paradigm is to not proceed to the next step until all your conditions are met:

```
print " Get ready to take off" 
v.mode = VehicleMode("AUTO")
v.armed= True
v.flush() # After this, write to vehicle has been made, but no guarantee change succeeded.
while not v.mode.name=='AUTO' and not v.armed==False and not api.exit:
    print " Waiting for AUTO and ARMED..."
    time.sleep(1)
print "Ready to take off"
```

Note: I will change our docs and examples to show the above.

There are two problems with this. Firstly it involves polling, which I object to on principle. Secondly, this will sit in a loop forever if the arming/mode change fails. This can happen because it is not possible to set modes/armed from certain states. We don't process the message that tells us if the other commands failed, so there is no way of determining what is going on.

So the question is, can/should we process command failure, and if so how?

The easiest way would probably be to make the flush() api synchronous and wait on responses to all preceding set messages (leaving exception if they fail). I wouldn't like to do it using callbacks because that is very messy code. We could just live with the code above.

@mrpollo - any ideas?

P.S. Don't know if it is possible in Python, but .NET does this quite nicely - you can make asynchronous requests and (at some point) wait on the result synchronously. This would allow us to write the code above without polling, and still return a value on failure. This is the problem with spending time in other languages, you learn about things you wish you had :-)

ggregory8:
Hi, I agree there should be some method to identify whether a command has failed.

Another common example is arming failing. As mentioned above an easy workaround is issuing the command then polling the state until it is set, an timeout would allow you to detect this has failed but is messy also. 

Are there currently any functions to handle command acknowledges?

zlite:
Adding Fredia,  the Droidplanner/Tower lead,  who has been thinking about
the same thing.

ne0fhyk:
[...]

PR #168, opened 2015-06-18
Added the ability to see command ack

djnugent:
A quick attempt at showing how I think command acknowledgment should be handled. This was denied as an option for issue #114 since mavlink is a "lossy protocol." I am not aware of the properties which make mavlink "lossy", but I feel this method would allow the user to see that the connection is acting "lossy" if the COMMAND_ACK message times out. 

Here is an example: https://gist.github.com/djnugent/edfbab1b2458d5773490

@mrpollo @billbonney thoughts?

### DIFF
--- a/droneapi/module/api.py
+++ b/droneapi/module/api.py
@@ -98,6 +98,24 @@ class MPCommandSequence(CommandSequence):
         self.__wp.wploader.add(cmd, comment = 'Added by DroneAPI')
         self.__module.vehicle.wpts_dirty = True
 
+    def command_ack(self, timeout = 0.5):
+        #wait indefinitely
+        if timeout is None: 
+            while self.__module.command_ack is None:
+                pass
+            result = self.__module.command_ack
+            self.__module.command_ack = None #reset cmd_ack
+            return result
+
+        #wait for timeout
+        start = time.time()
+        while (time.time() - start) < timeout and self.__module.command_ack is None:
+            pass
+        result = self.__module.command_ack
+        self.__module.command_ack = None #reset cmd_ack
+        return result
+
+
     @property
     def __wp(self):
         return self.__module.module('wp')
@@ -310,7 +328,6 @@ class APIThread(threading.Thread):
 
         # DroneAPI might generate many commands, which in turn generate ots of acks and status text, in the interest of speed we ignore processing those messages
         try:
-            self.module.mpstate.rx_blacklist.add('COMMAND_ACK')
             self.module.mpstate.rx_blacklist.add('STATUSTEXT')
         except:
             pass # Silently work with old mavproxies
@@ -329,7 +346,6 @@ class APIThread(threading.Thread):
             traceback.print_exc()
 
         try:
-            self.module.mpstate.rx_blacklist.remove('COMMAND_ACK')
             self.module.mpstate.rx_blacklist.remove('STATUSTEXT')
         except:
             pass # Silently work with old mavproxies
@@ -381,6 +397,8 @@ class APIModule(mp_module.MPModule):
         self.epv = None
         self.satellites_visible = None
         self.fix_type = None  # FIXME support multiple GPSs per vehicle - possibly by using componentId
+
+        self.command_ack = None
 
         self.next_thread_num = 0  # Monotonically increasing
         self.threads = {}  # A map from int ID to thread object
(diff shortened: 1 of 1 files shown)


eliao:
It seems problematic to block on the dronekit script's thread waiting on an ack.  Perhaps that works for some use cases, but it would not work for real time applications working over a lossy link.  In that case this would cause delays while waiting for an ack that will never come.  Over a lossy connection, one needs to assume that any and all packets can be dropped.

mrpollo:
I agree with Eric, the "lossy" really comes from the link between mavproxy and the autopilot, the transport can be anything from a 3DR Radio to Wifi, we can't control the time it takes to get an ACK, some of the symptoms are what @eliao describe, and yes all packets can be dropped.

@geeksville probably has something to say here too

djnugent:
I put a timeout in so we don't block the thread forever. The documentation would just address the downsides of waiting for a command ack and explain the timeout. And I feel you are less likely to drop a packet if the connection is hard wire serial. i.e. a companion computer to pixhawk which is going to be the main audience for dronekit-python. wait_valid() is a bigger sin with no timeout(unless downloading waypoints is different).  

Thanks for the input!

billbonney:
Checking the ACK form sending a command would confirm the command was accepted etc., but not necessarily acted upon. ie you could see an ACK to send a command to TAke Off, but then it won't Take Off as other conditions will not have been met. As I have written before, when issuing a command the only indication that it was accepted, is that the vehicle properties change to reflect that action.

billbonney:
PS: And qhat should be the value of the timeout? 0.5s or 5s or 10s. Timeouts are really tricky to gauge in protocols. A timeout for arming is actually about 5s and then variable.

djnugent:
Why would the vehicle not take off if it sent a MAV_RESULT_ACCEPT??? A MAV_RESULT_FAILED is sent if you are not in guided mode, not armed, and have already taken off a previously. I am confused...

djnugent:
Yes default timeouts are hard. We could also just set the default to 0 and make the user figure out the best timeout for their command.

tcr3dr:
Throwing in some Python comments here, if we're truly a Python lib, I expect we'd want a file descriptor you can poll for success, so it would easily integrate into gevent.

```
import gevent

def command_1():
   await_ack(task())

def command_2():
   await_ack(task())

gevent.spawn(command_1)
gevent.spawn(command_2)
# Nothing is blocked, it's just in the event loop
gevent.wait()
```

In my limited Python + async experience...

eliao:
Ideally, user code shouldn't have to deal with this.  In a non-lossy environment, users don't have to.  In a lossy environment, this could be solved at the protocol level via retries (preferred solution).  A somewhat worse solution is having DroneKit itself spawn a thread for these commands and ensuring they get completed.  This is what many of our applications currently do.

hamishwillee:
I'm most interested in the MAV_RESULT_FAILED as a way to understand there is a problem earlier than a timeout. The API should be asynchronous with some way to make it blockable if that is the way you want to code. It would be "best" if the callback is associated with a particular sent command and if for whatever reason it doesn't complete there is some cleanup on it.

hamishwillee:
Somewhat related, our solution should fit in with whatever we need to deliver #169

djnugent:
@tcr3dr I'm not familiar with events so you lost me there.
@eliao Retries on the lower lever is ideal, but that is beyond me. This was just an experiment.

I think call backs are great to prevent blocking but I imagine we are reaching out to "arduino" type developers who are coding amateurs. I bet they wouldn't care to wait for command_ack. It's just part of the API! The idea is to be able to do a,b,c,d... in sequential order. Dealing with mavlinks lossy connection is no different than dealing with a noisy sensor with arduino. You have to filter out bad behavior. This method is not super effective or ideal but it makes it easy for the new user.

Obviously we still need call back support for the advanced users.

So is this a good temporary solution or does it cause more problems than benefits?

billbonney:
The callback ack from the command long doesn't add much to the logic. Even for the advanced user. It's indicative in a log for example. But in the end most actions commanded are only relevant when you look at the result from the vehicle status. I.e. A change in mode by monitoring the heartbeat and the status text.  
If you want to make this easier for the Arduino crowd. Just make the default to ignore the ack as a false sense that a command has been sent and communicate in the docs how this works. Command long even has a bit to request the acknis even sent. 
In AP2 rarely do we check to see the result other than to put it in the log sometimes

djnugent:
Not a complete solution.

PR #170, opened 2015-06-24
Guided Mode Copter Topic

hamishwillee:
This adds a guided mode copter topic - possibly the most important copter topic of all. In order to develop this I significantly extended the guided-set-speed-yaw-demo.py example and associated example code documentation.

One thing I didn't do is rename the example source files, though I did give the example page a new title "Example: Guided Mode Movement and Commands (Copter)".

@mrpollo - worth moving the example and example docs to a new location to reflect the additional content?


hamishwillee:
@djnugent - Could you check this new content out? I am really happy with it, but it is a good idea for all new content to have an independent review.

djnugent:
@hamishwillee I'll give it a look!

djnugent:
@hamishwillee really impressed with this example!! It has a lot of great functionality. Some of it seems redundant, but I think that's the point if I am not mistaken; various ways to control the drones location?

djnugent:
@hamishwillee were you running v3.3 for this test? I'm not sure all of these actions work in v3.2.1. If that is the case then we shouldn't release this example until v3.3 is released.

hamishwillee:
> @hamishwillee really impressed with this example!! It has a lot of great functionality. Some of it seems redundant, but I think that's the point if I am not mistaken; various ways to control the drones location?

Glad you like it. Yes, some of the information is redundant. I have tried to be exhaustive in the example but in the guide cover the approaches you should use.  

Arguably the example should also guide users in the "best path" too - and I have tried to do that while still showing the less-relevant options. I thought it important to cover all the options because in theory the underlying commands allow quite different behaviour through the use of frames of reference. The fact that our implementation on Copter does not support all of these is useful to know (and it is also useful to provide an example which people can use to test this themselves - as I will when I got to testing this behaviour on Plane/Rover). In the interests of full disclosure, it is also this way because I didn't know anything when I started testing :-)

If you think I should remove these options then let me know.

> @hamishwillee were you running v3.3 for this test? I'm not sure all of these actions work in v3.2.1. If that is the case then we shouldn't release this example until v3.3 is released.

This was tested on 3.2.1. The main differences for 3.3 are in the behaviour of set_speed() - which I have documented. 3.3 also adds some changes for Camera, but since I couldn't test those I have omitted them from this example. This is an important enough use case that I hope to document it separately. 

Upshot, I think this can go out now/as soon as I've integrated fixes.

hamishwillee:
@djnugent I've added int he fix for the wrong goto commands in each section, and clarification in the example about why the two velocity commands are the same. Thank you very much for the review. Let me know if you think further changes are required following my responses above - otherwise this is good to merge.

djnugent:
@hamishwillee if it works on 3.2.1 then we are good :+1:. Just read the docs. They are very informative! If you fixed those errors than I'd say this is good to go. Definitely needs review by @mrpollo, this a large commit with stuff I am not 100% confident in. Again, good job :smile:

djnugent:
@rmackay9 not relevant to this PR, but what is going on in the second image of this commit? When the vehicle is using velocity commands it shows that zigzag patterns in the corners. Is that just PID tuning?

hamishwillee:
@djnugent @rmackay9 Possibly the effect of changing the yaw at the same time as the direction?

hamishwillee:
@mrpollo Hope you can find time to review/merge this today.

mrpollo:
all good stuff @hamishwillee, I think you cover a lot in here, we might want to keep an eye open for feedback on our guides as well

hamishwillee:
@mrpollo Thanks for pushing this in! Yes, we definitely need to keep an eye on feedback for this whole doc set. Perhaps actively seek it via survey or similar on the home page?

This topic is also the case where the new Page-TOC shines - without it finding relevant information in a [topic this big](http://python.dronekit.io/guide/copter/guided_mode.html) would be a lot more difficult.

I wonder if we need to add add a "jump to top" button because some topics are pretty long.

mrpollo:
@hamishwillee 
I think a message to encourage people to file bug claims on docs should be considered, "found a problem on the docs?, file here (link to gh issues)"

**Jump to top** ++ yes we need one

hamishwillee:
@mrpollo I added https://github.com/3drobotics/sphinx_3dr_theme/issues/35 for "jump to top". Lets discuss the "found problems on the docs" link in next team meeting?

PR #172, opened 2015-06-26
Add information about clearing ROI

hamishwillee:
Found information about clearing ROI in this post: https://github.com/diydrones/dronekit-python/issues/20#issuecomment-108656951

I tested this and it works. I didn't add test code demonstrating it. I could if this is considered useful?

### DIFF
--- a/docs/guide/copter/guided_mode.rst
+++ b/docs/guide/copter/guided_mode.rst
@@ -303,8 +303,11 @@ Send the `MAV_CMD_DO_SET_ROI <http://copter.ardupilot.com/common-mavlink-mission
         vehicle.send_mavlink(msg)
         vehicle.flush()
 
-From Copter 3.2.1 you can explicitly reset the ROI by sending the `MAV_CMD_DO_SET_ROI <http://copter.ardupilot.com/common-mavlink-mission-command-messages-mav_cmd/#mav_cmd_do_set_roi>`_ command with zero in all values. The vehicle will then face the direction of travel.
 
+.. versionadded:: Copter 3.2.1. You can explicitly reset the ROI by sending the 
+    `MAV_CMD_DO_SET_ROI <http://copter.ardupilot.com/common-mavlink-mission-command-messages-mav_cmd/#mav_cmd_do_set_roi>`_ 
+	command with zero in all parameters. The front of the vehicle will then follow the direction of travel.
+	
 The ROI (and yaw) is also reset when the mode, or the command used to control movement, is changed.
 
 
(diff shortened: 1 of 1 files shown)


djnugent:
@hamishwillee This isn't really a dronekit feature more of a arducopter/mavlink feature so I trust that it works. I have never needed to use it but I think is important we document it. 

@mrpollo +1

hamishwillee:
@tcr3dr This can be merged please. It is a minor docs enhancement.

docs/guide/copter/guided_mode.rst
```diff
@@ -303,9 +303,10 @@ Send the `MAV_CMD_DO_SET_ROI <http://copter.ardupilot.com/common-mavlink-mission
         vehicle.send_mavlink(msg)
         vehicle.flush()
 
-The ROI is "reset" when the mode or the method used to control movement is changed. If you've set a yaw or region-of-interest value then this will be set to the default (vehicle faces the direction of travel).
+From Copter 3.2.1 you can explicitly reset the ROI by sending the `MAV_CMD_DO_SET_ROI <http://copter.ardupilot.com/common-mavlink-mission-command-messages-mav_cmd/#mav_cmd_do_set_roi>`_ command with zero in all values. The vehicle will then face the direction of travel.
```
tcr3dr: This was still a bit unclear; I would propose text that addresses:

* "Available since version X" reads less ambiguously than "From", and is more familiar (i.e. Python docs)
* Is "all values" correct, as in lat/lon/alt also need to be 0? Is it more correct to say "all params"?
* "vehicle will then..." seems to need an active verb

And some proposed text:

```
*Available since Copter 3.2.1:* you can explicitly reset the ROI by sending the [`MAV_CMD_DO_SET_ROI`](http://copter.ardupilot.com/common-mavlink-mission-command-messages-mav_cmd/#mav_cmd_do_set_roi) command with zero in all parameters. The vehicle will return to face the direction of travel.
```

tcr3dr:
@hamishwillee Added a comment since I found the change a bit ambiguous. Review, change if desired, and I'll merge after your go-ahead.

hamishwillee:
@tcr3dr Thanks for the review - I find it hard to be objective about my text at the point I write it. 

I've now used the sphinx ".. versionadded::" directive which prefixes the text with "New in version" (I will get Kaitlyn to update the theme to highlight this like a note). I think using semantic markup in source is better because it allows for global change.

For the final sentence I changed it to "The front of the vehicle will then follow the direction of travel.". I think "follow" makes it clear that this isn't a one off change to face the direction of travel. 

Please merge if you're happy, and thanks again.

tcr3dr:
@hamishwillee This looks great! And I feel we'll be addressing version-specific features more as we move along, so a useful foray into it.

Issue #171, opened 2015-06-24
Takeoff command should check for NAN/Inf


hamishwillee:
``Vehicle.commands.takeoff`` could be called with a NAN or infinite value. We should probably check for those as either might result in a fly-away vehicle. 

```
    def takeoff(self, alt=None):
        if alt is not None:
            altitude = float(alt)
            self.__module.master.mav.command_long_send(self.__module.target_system,
                                                    self.__module.target_component,
                                                    mavutil.mavlink.MAV_CMD_NAV_TAKEOFF,
                                                    0, 0, 0, 0, 0, 0, 0,
                                                    altitude)
```

djnugent:
How does pymavlink handle NaN or infinity?

hamishwillee:
No idea! Apparently ArduPilot doesn't check this either (according to @billbonney). My take on this is that it would be good API design to raise an exception for invalid input if NaN or Inf appears here.

djnugent:
`veh.commands.takeoff(float("nan"))` is denied by arducopter
`veh.commands.takeoff(float("inf"))` crashed arducopter or SITL!!!!

billbonney:
I think in any case where you are converting a variable using float, you need to check for NaN or Inf. best of all would be check the value for altititude is within a set range in a global prefence, where alt isn't above the legal limit like 120m for example.

hamishwillee:
@billbonney Not sure about the last point - legal limits change, and so might a reasonable alt for a different vehicle (e.g. ArduRocket :-).

billbonney:
@hamishwillee Agreed limits change, but having a default to stop flyaway from a programming mistake is a good idea I think. it's a global setting that can be changed, but its there so when you do a calculation that is way off, your drone doesn't disapear is a Good Thing. Maybe even having it set to lower value would make sense. even at 30m it seems miles away ;)

djnugent:
Cant we just read the altitude from the fence altitude parameter and have that be the limit? And if it does not exist, then limit it to 400 ft.

hamishwillee:
I think we're all agreed that we should check for invalid input indicative of programmatic error (NaN or Inf). Re fly-away I think a check of some kind makes sense, but I'd argue that it should be in ArduPilot rather than DroneKit.

> And if it does not exist, then limit it to 400 ft.

IMO no - this should always be on some parameter rather than hard coded. What that parameter is would be decided by ArduPilot team, if they agree with this idea.

[...]

PR #174, opened 2015-06-26
Added inf and NaN check on takeoff

djnugent:
Solution for #171 
@mrpollo Does not check alt as instanceof  `float` because nan and inf are floats. It simply checks for nan/inf using the math lib


mrpollo:
:+1:

hamishwillee:
Excellent. Anywhere else in the API we should do this test?

PR #176, opened 2015-07-06
Add troubleshooting/restarting SITL sections

hamishwillee:
This adds some troubleshooting when using vagrant (it is possible for the image to corrupt) and also instructions on resuming/restarting).

In addition, it adds some minor fixes to rendering of paths (backslashes) and the bullet points.

### DIFF
--- a/docs/guide/sitl_setup.rst
+++ b/docs/guide/sitl_setup.rst
@@ -52,8 +52,8 @@ and Linux (though we recommend the native installations linked above).
 #. Install SSH (Windows only - SSH is present by default on Linux/Mac OSX)
 
    * Download and install `Git for Windows <https://git-scm.com/download/win>`_ (or another client that comes with SSH).
-     After installing you can locate the file using the command ``C:\where ssh`` (normally it is installed to **C:/Program Files (x86)/Git/bin/ssh.exe**
-   * Add the ssh.exe location to the *Path* (**System Properties | Advanced tab | Environment Variables | Path**)
+     After installing you can locate the file using the command ``where ssh`` (normally it is installed to **C:\\Program Files (x86)\\Git\\bin\\ssh.exe**
+   * Add the **ssh.exe** location to the *Path* (**System Properties | Advanced tab | Environment Variables | Path**)
 
 #. Create a new directory where you will run *Vagrant*, and open a command prompt/terminal in it: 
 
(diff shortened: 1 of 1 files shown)


tcr3dr:
At this point I'm comfortable floating [dronekit-sitl-runner](https://github.com/tcr3dr/dronekit-sitl-runner) as a cross-platform alterative to the Vagrant sections: https://github.com/tcr3dr/dronekit-sitl-runner

This uses precompiled binaries on S3 that are generated in a separate CI step: https://github.com/tcr3dr/dronekit-sitl-builder But all three versions of concern are supported and built for Windows, OS X, and Linux.

Basically, we should just outline what assurances are needed from that tool to use as a replacement for the Vagrant install (which I assume is less desirable in all circumstances).

mrpollo:
@tcr3dr if you feel like dksr (dk-sitl-runner) is the way to go we should remove all traces of sitl vagrant from the docs, lets create a new issue for this, and if this issue is now made irrelevant by that lets close this

long live [dronekit-sitl-runner](https://github.com/tcr3dr/dronekit-sitl-runner)!

hamishwillee:
@tcr3dr I can't get it to work on Windows (https://github.com/3drobotics/dronekit-sitl-runner/issues/9). 

Before replacing the vagrant stuff I would  hope that we could get this working on Plane and Rover (https://github.com/3drobotics/dronekit-sitl-runner/issues/3). 

Does this support "help" ie so people can get what versions of the binaries are available? 

Does this automatically set up any outputs or other defaults - ie what if you do "dronekit-sitl copter-3.4-dev"

hamishwillee:
@tcr3dr It would be best to push this out now, while we resolve issues with the sitl runner. The work is already done and minor.

docs/guide/sitl_setup.rst
```diff
@@ -52,7 +52,7 @@ and Linux (though we recommend the native installations linked above).
 #. Install SSH (Windows only - SSH is present by default on Linux/Mac OSX)
 
    * Download and install `Git for Windows <https://git-scm.com/download/win>`_ (or another client that comes with SSH).
-     After installing you can locate the file using the command ``C:\where ssh`` (normally it is installed to **C:\Program Files (x86)\Git\bin\ssh.exe**
+     After installing you can locate the file using the command ``C:\where ssh`` (normally it is installed to **C:/Program Files (x86)/Git/bin/ssh.exe**
```
tcr3dr: I'm confused about this change, shouldn't they be backslashes?

Also I believe that can read `where ssh` without the c:\

tcr3dr:
@hamishwillee Just one comment on the change, otherwise I'll merge.

hamishwillee:
@tcr3dr You're right - fixed by https://github.com/diydrones/dronekit-python/commit/6686cc07e743cbaf9ef71f16ffbfc0881f4c4dff . Note the double slashes are required because the first one is treated as an escape key.

hamishwillee:
@tcr3dr @mrpollo I'm closing this. Essentially we need to get DK-sitl running, and given our priorities this will never get in before that happens

Issue #181, opened 2015-07-07
Math module needs to be a dependency of dronekit


hamishwillee:
@mrpollo @djnugent I just tried the fix in #174 against our examples. They crashed because math is not defined in the example script. I don't believe that our examples should "have" to import this dependency, which means that you will need to import this in dronekit.

Crash is in the simple_goto.py example (unmodified).

```
if math.isnan(alt) or math.isinf(alt):
NameError: global name 'math' is not defined
```

Of course I could be doing something completely wrong ... in which case "help"!

djnugent:
@hamishwillee You are correct. I made an error with my PR. I will fix it.

tcr3dr:
Confirmed fixed. Added to v1.3.1 release

PR #182, opened 2015-07-07
Fixed import error for takeoff checks

djnugent:
Fixes issue #181


tcr3dr:
lgtm (tested and confirmed fix for #181)

Will be issuing a new point release.

hamishwillee:
@djnugent Thanks very much. Also works on the examples.

Issue #60, opened 2015-04-09
Attribute observer callbacks are called with heartbeat until disabled - after first called


eliao:
This seems odd to me, particularly because they are labeled "__on_change".  

eg:  if I do a 
vehicle.add_attribute_observer('mode', mode_callback)

mode_callback is called every heartbeat, regardless of mode change.

mrpollo:
https://github.com/diydrones/dronekit-python/blob/master/droneapi/module/api.py#L440

```python
        elif typ == "HEARTBEAT":
            self.__on_change('mode', 'armed')
```

not sure why we did this, @geeksville do you remember this?

geeksville:
Using heartbeat to detect mode changes is a good thing.  But yep, it would be much better to also keep a cache of mode and armed inside of the vehicle object and only publish mode and armed if the corresponding boolean changes state.

mrpollo:
perfect, I think this is what @eliao is calling **odd** here, I'll submit a PR with some caching, if the mode is still the same we won't publish ```__on_change('mode')``` same with **armed**

mhct:
@mrpollo Perhaps the cache should be done for all other properties as well, since there is no check for evaluating if they have changed or not.

hamishwillee:
@mrpollo @mhct Yes. The bug shows up on all properties not just mode.

hamishwillee:
@mrpollo FYI, it appears that the callback is called more often than the heartbeat -  when testing #184 I was  getting "lots" of records a second and the heartbeat only comes in once a second. I suspect this means that it is updated whenever a relevant message containing the attribute comes from the vehicle. That would certainly make more sense than updating on heartbeat. Can you confirm? (Note, I assumed this to be correct so that is what #184 states)

In  #184 I explain how you might cache the value and provide notification only when the value changes. I still think this caching should be done by our API.  

When do you think we might get some resource on this?

tcr3dr:
@hamishwillee I'll take over this issue. Since it's a significant departure from how callbacks are called so far in 1.x I think I'll reserve it for the next minor release, meaning I will focus effort on #184 first.

hamishwillee:
@tcr3dr Just to be precise:

- The way I *think* this works is that we publish observed values whenever messages come in with those values. The documentation delivered by #184 assumes this. Can you confirm?
- It could be that you are saying instead of the above we already cache the incoming values, and what is reported is the last received value, which is reported at some other schedule. If this is the case I'd say it is a bug

[...]

PR #184, opened 2015-07-08
Improve Rangefinder attributed documentation

hamishwillee:
1. Added vehicle.rangefinder attributed (only the class was documented). Updated class with units information and information about what happens if rangefinder not supported (returns None).

2. Updated the vehicle_state example and example docs to report the rangefinder values. The example docs now link to information on setting up simulated rangefinder. I took the option to change the Vehicle variable name from "v" to "vehicle" (purely for consistency with other examples) and tidy up the introductory text.

3. Updated the vehicle state guide to list the new attributes and show how they are obtained. I updated the section on using the observer to show how you can cache the value from the callback. I also made this behaviour sound less like a bug (#60).

### DIFF
--- a/docs/examples/vehicle_state.rst
+++ b/docs/examples/vehicle_state.rst
@@ -15,8 +15,11 @@ Running the example
 ===================
 
 The vehicle and DroneKit should be set up as described in :ref:`get-started`.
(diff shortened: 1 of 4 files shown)


tcr3dr:
As I mentioned in #60, I think documenting the current behavior is the right way to go. I'll follow up by reviewing this PR.

tcr3dr:
Merged via #237.

PR #199, opened 2015-07-14
Adds goto test.

tcr3dr:
Copied from vehicle_goto

Issue: timeout added to delay. I think this is because of requirements waiting

Also: rebased over #197

PR #208, opened 2015-07-16
Ensures observers can't be added twice. Fixes #110

### LINKED ISSUE #110: Vehicle notifier should only allow a callback to be added/removed once for a particular attribute
``Vehicle.add_attribute_observer`` will add an observer as many times as it is called, and  ``Vehicle.remove_attribute_observer`` will remove as many instances of the observer as times it is called.

IMO you might want to assign multiple observers to the same attribute, but I would expect them to be different. You might in theory want to also apply the same observer to multiple attributes. However a callback should really only be added once as an observer for an attribute, however many times it is called to be added.

The code below shows that if add_attribute_observer is called 4 times to add an observer, and then removed once, the callback is called 3 times. It should be called once, however many times it is added.

```
# Now change the vehicle into STABILIZE mode
v.mode = VehicleMode("STABILIZE")
	

# Define example callback for mode
def armed_callback(attribute):
    print "Armed-state changed: ", v.armed

v.add_attribute_observer('armed', armed_callback)
v.add_attribute_observer('armed', armed_callback)
v.add_attribute_observer('armed', armed_callback)
v.add_attribute_observer('armed', armed_callback)
v.remove_attribute_observer('armed', armed_callback)

print "Disarming..."
v.armed = False
v.flush()
```

tcr3dr:
Adds a fix for observers not being able to be added more than once.

Rebased over #199.

### DIFF
--- a/droneapi/lib/__init__.py
+++ b/droneapi/lib/__init__.py
@@ -351,7 +351,8 @@ class HasObservers(object):
         if l is None:
             l = []
             self.__observers[attr_name] = l
-        l.append(observer)
+        if not observer in l:
+            l.append(observer)
 
     def remove_attribute_observer(self, attr_name, observer):
         """
--- a/tests/sitl/test_110.py
+++ b/tests/sitl/test_110.py
@@ -1,0 +1,49 @@
+from droneapi.lib import VehicleMode
+from pymavlink import mavutil
+import time
+import sys
+import os
+
+def test_110(local_connect):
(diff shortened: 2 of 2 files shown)


tcr3dr:
Should not be merged until after #199.

mrpollo:
needs rebase?

tcr3dr:
Closed to change branch. https://github.com/diydrones/dronekit-python/pull/224

PR #212, opened 2015-07-17
Adds CONTRIBUTING info for tests.

tcr3dr:
Needs work.

### DIFF
--- a/docs/contributing/contributions_api.rst
+++ b/docs/contributing/contributions_api.rst
@@ -30,21 +30,63 @@ repository and contribute changes back to the project master branch using pull r
 Test code
 =========
 
-Test code should be used to verify new and changed functionality. Changes to DroneKit can be tested locally by 
-rebuilding your git branch and running it against SITL. The links below provide information on how 
-to set up a development environment on your development computer:
+Test code should be used to verify new and changed functionality. Changes to DroneKit can be tested locally by rebuilding your git branch and running it locally. The links below provide information on how to set up a development environment on your development computer:
 
 * :ref:`dronekit_development_linux`
 * :ref:`dronekit_development_windows`
 
-.. note:: At time of writing (June 2015) we do not yet have formal acceptance tests (unit tests, system tests).
+Running Web Client Tests
+========================
 
-We recommend that you provide your test script as a gist, and run any the 
-:ref:`DroneKit-Python Examples <example-toc>` that are relevant.
- 
-For example, the addition of a new API item for retrieving a vehicle attribute would need test code for 
-getting, setting (if applicable) and printing the attribute. An easy way to create this would be to 
-update and run :ref:`example-vehicle-state`.
+Web client tests use nose. First export a DRONEAPI_KEY in the format `<id>.<key>`. The run:
+
+.. code:: bash
+
+    cd dronekit-python
+    nosetests tests/web
+
+Running Integrated Tests
+========================
+
+Integrated tests use the SITL environment to run DroneKit tests against a simulated copter.
 
 
+.. code:: bash
 
+    cd dronekit-python/tests
+    python -um sitl
+
+You can configure any of these environment variables to update it:
+
+#. TEST_SPEEDUP - Speedup factor to SITL.
+#. TEST_RATE - Sets framerate. (Default is 200 for copter, 50 for rover, 200 for plane.)
+#. TEST_RETRY - Retry failed tests. This is useful if your testing environment performs inconsistently with the simulated copter.
+
+You can also specify specific test names using:
+
+.. code:: bash
+
+    python -um sitl test_1.py test2_.py ...
+
+Writing a new Test
+
+You can write a new integrated test by adding a file with the naming scheme `test_XXX.py` to the `sitl` directory. In this file, functions with the prefix `test_` will be called with the `local_connect` parameter. For example:
+
+.. code:: python
+
+	from testlib import assert_equals
+
+	def test_parameters(local_connect):
+	    v = local_connect().get_vehicles()[0]
+
+	    # Simple parameter checks
+	    assert_equals(type(v.parameters['THR_MIN']), float)
+
+This checks to see that the parameter object is of type float. Use assertions to test your code is consistent. Avoiding printing any data from your test.
+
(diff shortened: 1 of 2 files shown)


docs/contributing/contributions_api.rst
```diff
@@ -30,21 +30,63 @@ repository and contribute changes back to the project master branch using pull r
...
+Running Web Client Tests
+========================
 
-We recommend that you provide your test script as a gist, and run any the 
-:ref:`DroneKit-Python Examples <example-toc>` that are relevant.
- 
-For example, the addition of a new API item for retrieving a vehicle attribute would need test code for 
-getting, setting (if applicable) and printing the attribute. An easy way to create this would be to 
-update and run :ref:`example-vehicle-state`.
+Web client tests use nose. First export a DRONEAPI_KEY in the format `<id>.<key>`. The run:
```
hamishwillee: What is nose? Should be a link. Similarly, the DRONEAPI_KEY should link to our cloud docs.

docs/contributing/contributions_api.rst
```diff
@@ -30,21 +30,63 @@ repository and contribute changes back to the project master branch using pull r
...
 
+.. code:: bash
 
+    cd dronekit-python
+    nosetests tests/web
 
+Running Integrated Tests
+========================
+
+Integrated tests use the SITL environment to run DroneKit tests against a simulated copter.
```
hamishwillee: Again, worth a sentence to provide high level info about what our integrated tests are/how they differ from the web tests.

So basically there is a script in **dronekit-python/tests** called sitl that you run using ``python -um sitl`` to start sitl? Is this the sitl runner based?

docs/contributing/contributions_api.rst
```diff
@@ -30,21 +30,63 @@ repository and contribute changes back to the project master branch using pull r
...
+
+
+.. code:: bash
+
+    cd dronekit-python/tests
+    python -um sitl
+
+You can configure any of these environment variables to update it:
```
hamishwillee: This works on all platforms - wndows/mac/linux? ie whatever the platform environment variable approach is used that will be used?

docs/contributing/contributions_api.rst
```diff
@@ -30,21 +30,63 @@ repository and contribute changes back to the project master branch using pull r
...
+#. TEST_RATE - Sets framerate. (Default is 200 for copter, 50 for rover, 200 for plane.)
+#. TEST_RETRY - Retry failed tests. This is useful if your testing environment performs inconsistently with the simulated copter.
+
+You can also specify specific test names using:
+
+.. code:: bash
+
+    python -um sitl test_1.py test2_.py ...
+
+Writing a new Test
```
hamishwillee: This should be a heading.

hamishwillee:
Need to add information about the different classes of tests and how to write for each of them as per #223

tcr3dr:
@hamishwillee I updated this doc with all suites of tests and addressed your comments. Feel free to review again.

These docs will accompany the 1.4.0 release (which is mostly just the addition of the test infrastructure).

hamishwillee:
Hi @tcr3dr

Look pretty good. I pushed some minor subedits for style structure.  Please check you're happy with those.

Comments:

1. This branch has no unit tests so I can't assess how useful the unit test advice is
1. FYI, the theme I'm testing against doesn't differentiate 4th level headings, which is one reason I changed the heading levels (the other being I think it improves the structure).
1. The web client is not actually part of the Python public API - as you will see in the API reference it is not documented. It would be, but too much was not implemented so we hid it. We might want to remove this altogether from docs until next version of the cloud API. 
1. The integration test advice says ""Avoiding printing any data from your test.". However test_12.py file we supply does have prints. I think this is an old version of test_simple.py and probably should be removed
1. We advise to use asserts. a) should we state in text that you can pull asserts_equals from testlib? b) That is the only assert I see - do we need to provide (and document) others or is this enough?
1. Should we be further recommend need for meaningful names for the files. test_simple.py is not as useful for searching as test_vehicle_parameters.py (and I think we should change this name too unless this is just considered a "demo" file - in which case perhaps "test_simple_demo" might be even better.)
1. Should we also recommend that each file and method added be given brief documentatio? I think this a good idea.
1. We might want to move the example of setting environment variables (or copy it) up to the integration test section, since here we also state that you need to set them.

tcr3dr:
@hamishwillee I address most of your concerns.

* The branch is now rebased so you can test if unit tests work. 
* I added a warning to, but didn't remove, the Web API test section (for completeness)
* The recommendations for `test_simple` and `test_12` are good and I'll open an issue for those.

If this passes CI and you are happy with it, let me know!

hamishwillee:
@tcr3dr Really good fix/response! I created #233 (like #232) and added a couple of minor typo fixes to highlighting of assert_equals.

My only concern is that we refer to mock but there is no useful real-world example (e.g. test_api.py) that shows enough to know "correct use".

Consider the example:

```
import mock
from mock import MagicMock
import droneapi
from droneapi.module.api import APIModule
from nose.tools import assert_equals

def test_mode():
    api = APIModule(MagicMock())
    res = api.get_connection()
    assert_equals(len(res.get_vehicles()), 1)
```
From a user perspective trying to work out how to use Mock with dronekit I wonder what APIModule is (since it isn't public). Why not use the local_connection? 
Reading the mock docs I might assume that here you are providing the connection to use, and this might set up a fake connection definition that guarantees that get_vehicles() returns 1, but I can't see where that might be set?
The use of Mock implies that these tests never run against a real device - correct? 

Upshot, a little more explanation of mock or a better example which we link to as an example would help. 

I would be happy for what we have here now to go out and this additional material to be provided in a separate issue if you preferred.

tcr3dr:
@hamishwillee Good points. I'll split that into a separate issue. Mostly because I'm not super familiar with Mock myself yet, though I've seen it used extensively by @eliao to good effect. As the tests improve, so shall the docs.

eliao:
I'm relatively new to Mock myself, but it seems quite handy.  It's very useful for isolating tests so that they test only exactly what they are meant to.  If the tests include networking, etc, then they can fail/succeed due to things that are outside their control.  Unit testing vs. functional testing.

PR #211, opened 2015-07-17
Improve docs to reduce issues from #132

hamishwillee:
This adds explicit warnings to the docs that you need to call flush after clear. We might remove this warning after #132 is fixed.

It also updates the examples appropriately. Those changes can be left after #132 is fixed because they "do no harm"

Issue #132, opened 2015-05-20
Vehicle.commands is throwing away the first command sent


hamishwillee:
The first command set in the mission is being discarded (probably by DroneKit). The following code shows how to create  a number of missions, starting with takeoff.

```
#add a takeoff command
altitude=100 #target altitude
pitch=45 #take off pitch. Need to check if degrees or radians, and what is a reasonable valued.
cmd = Command(  0, 0, 0, mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT, 
                mavutil.mavlink.MAV_CMD_NAV_TAKEOFF, 0, 0, 
                pitch, 0, 0, 0, 0, 0, altitude)
cmds.add(cmd)
			   
lat = -35.363768
lon = 149.154275
altitude = 110
cmd = Command(0,0,0, mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT, mavutil.mavlink.MAV_CMD_NAV_WAYPOINT,
    0, 0, 0, 0, 0, 0,
    lat, lon, altitude)
cmds.add(cmd)

lat = -35.3666207
lon = 149.1662800
altitude = 120
cmd = Command(0,0,0, mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT, mavutil.mavlink.MAV_CMD_NAV_WAYPOINT,
    0, 0, 0, 0, 0, 0,
    lat, lon, altitude)
cmds.add(cmd)

lat = -35.363769
lon = 149.154276
altitude = 130
cmd = Command(0,0,0, mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT, mavutil.mavlink.MAV_CMD_NAV_LOITER_UNLIM,
    0, 0, 0, 0, 0, 0,
    lat, lon, altitude)

cmds.add(cmd)
v.flush()
```

The output below shows the missions as sent, and the missions that are then re-retrieved from the vehicle. The first mission item is simply discarded (the first mission item is now the home)

```
AUTO> Create mission
Got MAVLink msg: MISSION_ACK {target_system : 255, target_component : 0, type : 1}
Sent waypoint 0 : MISSION_ITEM {target_system : 1, target_component : 0, seq : 0, frame : 3, command : 22, current : 0, autocontinue : 0, param1 : 45, param2 : 0, param3 : 0, param4 : 0, x : 0, y : 0, z : 100}
[...]

PR #213, opened 2015-07-17
Improve docs to reduce issues from #132 mk 2

hamishwillee:
This adds explicit warnings to the docs that you need to call flush after clear. We might remove this warning after #132 is fixed.

It also updates the examples appropriately. Those changes can be left after #132 is fixed because they "do no harm"

This replaces #211


tcr3dr:
lgtm!

PR #216, opened 2015-07-17
Add python-dev as a dronekit dependency on linux

hamishwillee:
Missing dependency. Thought I'd already added it but can't find a matching pull request.


mrpollo:
:+1:

PR #161, opened 2015-06-13
Created windows installer. Use batch file to run it

stephendade:
Installer details
===============

> Added by HamishW - from Stephen email

1. The installer looks for the MAVProxy directory and installs the contents of the ./droneapi and ./examples directories into there.
1. Then simply use the "module load droneapi.module.api" to load it in MAVProxy. (#HW note, it does use the init script too!)
1. The examples can be loaded by the usual "api start ./examples/folder/script"
1. A minimum of MAVProxy 1.4.20 is required to run DroneAPI

Using Installer script
=================

Preconditions
* Install [Inno Setup](http://www.jrsoftware.org/isinfo.php) installed at default location. 
* The script assumes Python is installed in C:\Python27

Build the installation file using \dronekit-python\windows>droneapiWinBuild.bat

The script gets the version number from dronekit's setup.py (It assumes there is a line like ``# version = "12344"``)

PR #217, opened 2015-07-18
Windows Installer: Added protobuf to installer

stephendade:
Protobuf is required by dronekit-python, but is not always included in a MAVProxy installation. Thus this patch gathers up protobuf into the dronekit-python installer.

More background:
MAVProxy is dependant on the mavlink package. Mavlink lists protobuf as "optional", so it may or may not exist in the user's environment.
When MAVProxy is compiled into a setup.exe, protobuf may or may not be included due to the above. Thus we can't assume that it's always included in a MAVProxy environment.

### DIFF
--- a/windows/droneapiWinBuild.bat
+++ b/windows/droneapiWinBuild.bat
@@ -9,6 +9,29 @@ for /f "tokens=*" %%a in (
  set VERSION=%%a
  )
 
+rem -----Get protobuf and unzip-----
+C:\python27\scripts\pip.exe install --download .\ protobuf
+rem get the protobuf fullname
+for /f "tokens=*" %%a in (
+ 'dir /B protobuf-*'
+ ) do (
+ set PROTOBUFFULL=%%a
+ )
+
+rem Use 7zip to unzip files. Works for both 32 and 64 bit versions
+"C:\Program Files\7-zip\7z.exe" x -y %PROTOBUFFULL%
+"C:\Program Files\7-zip\7z.exe" x -y .\dist\protobuf*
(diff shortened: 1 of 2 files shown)


stephendade:
@hamishwillee and @djnugent This leads on from the issues seen in #161

hamishwillee:
@stephendade - thanks for the explanation - sounds very sensible. @tcr3dr What do we need to get this fix merged?

tcr3dr:
Thanks @stephendade! I'll steward this. I want to modify CI to reproduce the problem (somehow it's not), then have this patch fix it. Then we're good to merge.

(@hamishwillee As for the "travis failing" issue, incidentally, creating a separate branch consisting of these commits, then merging it, will mark this PR as merged. So this is the temporary permissions workaround.)

hamishwillee:
@tcr3dr Thanks for the explanation. I really should take some time to understand Travis.

tcr3dr:
@stephendade I've succeeded in using the installer on Windows, but I had to modify the .bat script to remove the `-compile` flag (not its argument, just that string) from invoking ISCC.exe. Can you confirm that your batch script still creates the installer even without the -compile flag?

From what I've seen, the `-compile` flag is unnecessary or undocumented: http://www.jrsoftware.org/ishelp/topic_compilercmdline.htm

stephendade:
@tcr3dr Yep - it works fine without the "-compile" option. Honestly can't remember why I added it in the first place :)

tcr3dr:
Merged as of #236. Thanks @stephendade!

PR #243, opened 2015-07-27
v1.4.1

tcr3dr:
See release notes for details.


hamishwillee:
@tcr3dr Any idea when this likely to go live? 2 hours is a long wait isn't it?

tcr3dr:
@hamishwillee Should have worked... https://pypi.python.org/pypi/droneapi/1.4.1

hamishwillee:
Ah, I mean that the docs build is still on 1.3.1 - this is just PIP? http://python.dronekit.io/about/release_notes.html

tcr3dr:
Got it, should be updated now. http://python.dronekit.io/about/release_notes.html

hamishwillee:
Yes- updated thanks:

* Did anything have to be pushed or was it just a matter of waiting? I ask so that I know how long I should be patient for before nagging.
* I'm having to refresh my local cache on every page to get these to display - can we somehow force a cache refresh?

tcr3dr:
@hamishwillee That was all on me, I forgot that step. And It's just a one-liner, so not a problem at all to ping me on it.

Let me look into cache refreshing, I need to familiarize myself with the hosting we use...

hamishwillee:
@tcr3dr Not a worry - but you'll never get back that carefully cultivated aura of infallibility :-) Thanks for the information.

tcr3dr:
The reality:

hamishwillee:
+1 :-)

Issue #163, opened 2015-06-16
DK Build Toolchain updates for Windows DroneKit Installer


hamishwillee:
Update our build system to automatically create new Windows DroneKit installation packages on every rebuild. This includes:

- Establish process for Windows versioning and distribution
- CI/CD builds for Windows on each release http://www.appveyor.com/

This needs to be addressed when #161 is merged.

tcr3dr:
This is actually doable now that Appveyor can run the installer. See #236.

I suppose we should publish builds on each tag release. This shouldn't be too hard to setup. @hamishwillee, should this happen before adding Windows Installer instructions to the tutorial?

hamishwillee:
@tcr3dr Yes please. I can't update the getting started to include the windows installer option until I know where the installers will be hosted (ideally with a permanent "latest version" link). That is part of what this issue needs to deliver.

tcr3dr:
@hamishwillee We have our first Windows builds:

* https://s3.amazonaws.com/dronekit-installers/windows/dronekit-windows-latest.exe
* https://s3.amazonaws.com/dronekit-installers/windows/dronekit-windows-v1.4.2.exe

These will continue to be built as we publish releases. Feel free to link to them!

tcr3dr:
@hamishwillee Feel free to close this issue in favor of #164.

hamishwillee:
Closed in favour of #164

PR #253, opened 2015-07-28
Publishes Windows installers on versioned tags.

tcr3dr:
See #163. This publishes to AWS any tagged (vX.X.X) releases, as well as publishes a "latest" build.

@mrpollo Can you set up a cloudfront link?


mrpollo:
lgtm

Issue #189, opened 2015-07-09
Remove older, unused files.


tcr3dr:
It seems these are old and could be removed:

* [x] https://github.com/diydrones/dronekit-python/blob/master/scripts/mavgateway
* [x] https://github.com/diydrones/dronekit-python/blob/master/ELC.md
* [x] https://github.com/diydrones/dronekit-python/blob/master/compile-protobufs.sh

tcr3dr:
Completed.

PR #259, opened 2015-07-30
Consolidates scripts into scripts/ directory.

tcr3dr:
These "maintenance" files were scattered all over the place, so I think they should all belong under the scripts/ directory. 

Also this removes the old "mavgateway" script, see #189.

Nominating for [minor]


tcr3dr:
I believe @hamishwillee and @mrpollo, you should both review.

hamishwillee:
@tcr3dr  update to scripts/get_release_notes.py looks sane to me and I assume you have tested it. I don't have a view/experience on the other files.

mrpollo:
quickly glanced through this, lgtm

Issue #251, opened 2015-07-28
DRONEKIT_PATH variable


djnugent:
Currently dronekit only supports relative and absolute paths. It would be nice if we could point `api start <file>` to look at a set of directories when running a script. DRONEKIT_PATH would point to your project folders.

This idea was brought up in a recent dronekit session Chris and I had. At a minimum we should specify in the documentation that if that user can't specify the directory in which mavproxy is started(i.e. running SITL), then they must start use an absolute path to call their script.

DRONEKIT_PATH should not be necessary for when we run dronekit without mavproxy. But at a minimum we should document that the user should provide an absolute path when running dronekit scripts.

zlite:
Also, let's update the documentation to show people how to download the
example files:

git clone https://github.com/dronekit/dronekit-python/examples

On Tue, Jul 28, 2015 at 2:28 PM, Daniel Nugent <notifications@github.com>
wrote:

> Currently dronekit only supports relative and absolute paths. It would be
> nice if we could point api start <file> to look at a set of directories
> when running a script. DRONEKIT_PATH would point to your project folders.
>
> This idea was brought up in a recent dronekit session Chris and I had. At
> a minimum we should specify in the documentation that if that user can't
> specify the directory in which mavproxy is started(i.e. running SITL), then
> they must start use an absolute path to call their script.
>
> DRONEKIT_PATH should not be necessary for when we run dronekit without
> mavproxy. But at a minimum we should document that the user should provide
> an absolute path when running dronekit scripts.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dronekit/dronekit-python/issues/251>.
>



-- 
Chris Anderson
CEO, 3D Robotics
chris@3drobotics.com

hamishwillee:
Yes, it is painful to have to enter full search paths - I'm always thinking "what is my current path" and I tend to just close the prompt and reopen it in the new folder I want to run. I like the idea, but I don't think we'll need it when we have dronekit running as a module, so I guess whether we do this depends on priority/effort vs how quickly we'll get dronekit running as an independent module. @tcr3dr do you have a view on that?

@djnugent  @zlite Regarding the docs - In both cases the "missing" documentation is present (see below).

> But at a minimum we should document that the user should provide an absolute path when running dronekit scripts.

We do document this [here in the getting started](http://python.dronekit.io/guide/getting_started.html#running-an-app-example):
[...]

PR #261, opened 2015-07-30
Update gettingstarted/running examples

hamishwillee:
This request addresses the documentation concerns raised in #251

Specifically it makes clear how to download and run the examples in both getting started and running the examples sections. 

The sections are now very similar - the difference being that the getting stated reflects a specific example, while the other is more general.  I think they are both more clear now.

### DIFF
--- a/docs/examples/running_examples.rst
+++ b/docs/examples/running_examples.rst
@@ -1,24 +1,53 @@
+.. _running_examples_top:
+
 ====================
 Running the Examples
 ====================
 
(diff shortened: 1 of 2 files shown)


mrpollo:
@tcr3dr I believe Travis choked on a false/positive, can you check that out plz?

tcr3dr:
@mrpollo Confirm, it's just the test fixture. Will go away with v2

tcr3dr:
Merged via #264 (added another small commit on top)

Issue #267, opened 2015-08-03
Add loading of dronekit module to mavinit.scr in dronekit installer


hamishwillee:
It would remove one step for users if the dronekit installer automatically set up dronekit api to be loaded into mavproxy. This is particularly important because the mavinit.scr is cleared whenever you reinstall mavproxy.

The required line is (I believe):

```
echo module load droneapi.module.api >> %HOMEPATH%\AppData\Local\MAVProxy\mavinit.scr
```

Ideally run this conditional on it not being present already :-)

@stephendade @tcr3dr  Thoughts?

stephendade:
@hamishwillee - done in #271 :)

hamishwillee:
@stephendade Thank you! I'll update the docs in line with this today (hopefully!)

tcr3dr:
We will no longer need this since DKPY2.

PR #268, opened 2015-08-03
Add all changes needed to support windows dk installer

hamishwillee:
This pull addresses #264.

This updates the getting started to use Windows DK installer rather than the WinPython installation instructions. Unfortunately the windows installer is started differently, so there are flow on changes to:

* how the connection is set up. 
* how to start a script (since you can't start MAVProxy in a specific directory on Windows, I've had to change the instructions to "first" suggest the full path.
* Where the examples are (add note that these are available on windows from the installer). 

Similar changes needed to be made on the How to Run the examples page.

Last of all, the "Building DroneKit-Python on Windows" was updated to include the setup instructions formerly in Getting Started. I also added the additional dependencies reported as missing on Stack overflow.

There might need to be minor updates to this as a result of #267.

I also based this on patch. Hope that is the correct thing to do.

### DIFF
--- a/docs/contributing/developer_setup_windows.rst
+++ b/docs/contributing/developer_setup_windows.rst
@@ -4,11 +4,63 @@
 Building DroneKit-Python on Windows
 ===================================
 
-The setup for *developing* DroneKit-Python on Windows is almost the same as for *using* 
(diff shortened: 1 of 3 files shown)


tcr3dr:
Merged by #273.

PR #273, opened 2015-08-06
Add all changes needed to support windows dk installer.

tcr3dr:
See #268.

Rebases onto master. Re-runs tests.

PR #275, opened 2015-08-07
Update docs in line with #273

hamishwillee:
This removes the statement that you still have to load dronekit into MAVProxy, and add another saying that the **latest** version sets up MAVProxy to load DroneKit. 

We don't have a link (yet) for the whole set of dk-installer, which is a pity. But it does mean that using "latest" means I don't need to be specific about what version the change to installer version went in.


hamishwillee:
@tcr3dr @mrpollo This is now very out of date and not relevant to DKPY2. CLosing

PR #279, opened 2015-08-10
v1.4.3

tcr3dr:


hamishwillee:
@tcr3dr I just pulled down the dronekit windows installer latest release link to try this out and it downloaded what asserts to be 1.4.2. Can you confirm that the dk installer should have been rebuilt and if it wasn't what happened?

Issue #278, opened 2015-08-10
VehicleMode should be easy to compare for equality


atomictom:
VehicleMode should be an enum or at least enum-like. The biggest issue with the current implementation is:
```python
VehicleMode("GUIDED") == VehicleMode("GUIDED") is False
```
This means that determining which mode you are in either requires you to make your own enum:
```python
GUIDED = VehicleMode("GUIDED")
LAND = VehicleMode("LAND")
```
and only use those variables. 

Test for string equality:
```python
str(vehicle.mode) == "VehicleMode:GUIDED"`
````
Or access the `name` field (which is not documented):
```python
vehicle.mode.name == "GUIDED"
```

Additionally, being an enum would allow for code that checks whether a given mode exists programmatically:
```python
if "LAND" not in VehicleMode:
    # do something
```
and to iterate through the available enums:
```python
for mode in VehicleMode:
    menu.add(mode)
```

The type of enum I have in mind is detailed in [this StackOverflow post](http://stackoverflow.com/questions/36932/how-can-i-represent-an-enum-in-python)

hamishwillee:
Hi @atomictom 

I absolutely agree that we should be easy to test for equality, to iterate through the modes, and to test for existence as you do above. I'm not sure that an enum is the right way to do this - probably this should be handled similarly to parameters (except they need to be observable too). 

I have therefore changed the title of this issue.

Regards
Hamish

P.S. [VehicleMode.name](http://python.dronekit.io/automodule.html#droneapi.lib.VehicleMode.name) is documented, and this would be the right way for comparing for equality.

atomictom:
An enum was the first thing that came to mind, but anything that filled those requirements would work fine. One additional thought I had was, it would be nice if there were only one reference to a VehicleMode. That way, you could test for equality with the `is` operator, perhaps like so:
```python
if vehicle.mode is VehicleMode.GUIDED:
    # do stuff
```

Also, you're right! Perhaps if there was a code example in the docs that used it (perhaps in an if-statement):
```python
# Test to see if we are in GUIDED mode
if vehicle.mode.name == "GUIDED":
    vehicle.armed = True
```
Thanks,
Thomas

hamishwillee:
@atomictom The code examples do use it - see [Set attributes](http://python.dronekit.io/guide/vehicle_state_and_parameters.html#vehicle-state-set-attributes). 

[...]

PR #288, opened 2015-08-14
Tmp modecompare 278

hamishwillee:
Add explicit text that you need to compare against **Vehicle.mode.name** (ie you can't compare ``Vehicle.mode == VehicleMode('GUIDED')`` ). You can however compare ``Vehicle.mode.name == VehicleMode('GUIDED').name``

I've put this in the API reference for the Mode because we don't have a better place. I made a conscious decision to not document attributes separately in the guide, but instead to put documentation in the class (if there is one) and the Vehicle attribute itself if there isn't an associated class. 

This was to address the specific comment https://github.com/dronekit/dronekit-python/issues/278#issuecomment-130138613

### DIFF
--- a/droneapi/lib/__init__.py
+++ b/droneapi/lib/__init__.py
@@ -247,6 +247,17 @@ class VehicleMode(object):
 
         # Set the vehicle into auto mode
         vehicle.mode = VehicleMode("AUTO")
+        vehicle.flush()  # Flush to guarantee the write is complete
+
+    After setting the mode you can test that it has changed by comparing the ``name`` member:
+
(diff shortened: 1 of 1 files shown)


tcr3dr:
@hamishwillee Should this be rewritten as per https://github.com/dronekit/dronekit-python/pull/292 to indicate equality comparisons?

You can push to the other branch the docs changes if you'd like and close this PR.

hamishwillee:
@tcr3dr Suggest you just merge #292 and close this. I will add needs docs labels to #292 and think about what to add (if anything) once it is in.

PR #292, opened 2015-08-14
VehicleMode is now comparable

### LINKED ISSUE #278: VehicleMode should be easy to compare for equality

mrpollo:
this change makes it safe to compare `VehicleMode` objects with other `VehicleMode` objects and with strings

```python
vehicle.mode == VehicleMode('GUIDED')
vehicle.mode == 'GUIDED' 
```

```python
vehicle.mode != VehicleMode('GUIDED')
vehicle.mode != 'GUIDED' 
```

fixes #278

### DIFF
--- a/dronekit/lib/__init__.py
+++ b/dronekit/lib/__init__.py
@@ -285,6 +285,12 @@ class VehicleMode(object):
     def __str__(self):
         return "VehicleMode:%s" % self.name
 
+    def __eq__(self, other):
+        return self.name == other
+
+    def __ne__(self, other):
+        return self.name != other
+
 class APIConnection(object):
     """
     An API provider.
@@ -524,7 +530,7 @@ class Vehicle(HasObservers):
 
     .. py:attribute:: rangefinder
 
-        :py:class:`Rangefinder` distance and voltage values. 
+        :py:class:`Rangefinder` distance and voltage values.
 
 
     .. py:attribute:: channel_override
@@ -750,7 +756,7 @@ class Vehicle(HasObservers):
             vehicle.set_mavlink_callback(mavrx_debug_handler)
 
         :param callback: The callback function to be invoked when a raw MAVLink message is received.
-		
+
         """
         self.mavrx_callback = callback
 
--- a/tests/unit/test_api.py
+++ b/tests/unit/test_api.py
@@ -1,9 +1,18 @@
 import mock
 from mock import MagicMock
-import dronekit
-from dronekit import FakeAPI
+from nose.tools import eq_, assert_not_equal
+import droneapi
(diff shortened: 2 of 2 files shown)


hamishwillee:
@mrpollo Wouldnt it make more sense just to turn vehicle.mode into a string? That has all the comparison operators you'd ever need :-) 

I don't see any benefit in having a class (other than compatibility). Also we will probably want to add API to get list of modes at some point, and if we do that, having as a string would be even better for comparing.

mrpollo:
old...

@hamishwillee I agree on the sentiment, but it's how we have been doing things in the past, "consistency"?, also we could easily add a list of modes to this object, and that timestamp

@tcr3dr given the recent changes, whats the best way to proceed with this and all the older prs?

hamishwillee:
Probably easier to create a new PR than merge some of these now? You guys probably have a better idea than me the costs/benefits of a particular implementation - IMO it doesn't matter so much which we use, but this is a very good time for us to introduce API compatibility breaks.

tcr3dr:
Chewing through old PRs can just happen case by case.

What other attributes would VehicleMode object need?

* `index` for the underlying mode index
* `timestamp`, though I feel it's weird to attach that onto the property?

I'm in favor of making `.mode` settable by a string, but if there's reason to expose more data we should keep it exposed as a VehicleMode class. Anywhere it can be cast to a string it will be, so I'm not too concerned at simplifying something may actually become more complex over time.

mrpollo:
lets keep it as a class for consistency with all the other properties we already expose as classes, ardupilot is full of surprises, who knows what comes down the road that we might need to expose

hamishwillee:
OK. 
Looks like you're going to have to rebase this.

Issue #285, opened 2015-08-13
condition_yaw doesn't work without a call to send_global_velocity


atomictom:
I was testing out the `condition_yaw` function in the docs and while the command would send a MAVLink message, and the drone (SITL) would respond with success, the drone would not yaw (as indicated by the `vehicle.attitude.yaw` field). However, if I precede the `condition_yaw` call with a `send_global_velocity(0, 0, 0)` call (again, taken from the docs), the drone will rotate successfully. 

I am executing these commands in GUIDED mode. As far as I can tell, there doesn't seem to be a mention of this requirement in the docs (I guessed at it since the yaw examples included `send_global_velocity`).

hamishwillee:
@atomictom Can you create a gist of your example code please - ideally based off http://python.dronekit.io/examples/guided-set-speed-yaw-demo.html ?

atomictom:
https://gist.github.com/atomictom/9bae5cb7642045b184da

It uses mostly unmodified versions of the `condition_yaw` and `send_global_velocity` functions. If you uncomment either of the lines with `# send_global_velocity(0, 0, 0)` then it yaws, otherwise it won't.

I load it up using `mavproxy.py --master=172.28.128.1:5760 --cmd="module load droneapi.module.api ; api start gist.py"` connecting to a vagrant version of SITL. After you see the confirmation `Got MAVLink msg: COMMAND_ACK {command : 115, result : 0}` the yaw should (when working correctly) change significantly.

hamishwillee:
The gist didn't work (missing yaw() ) but I modified it and yes, I think you are correct. You can of coruse just move to your current position, but it is a bit annoying. This will be autopilot behaviour, not dronekit as such, but yes we should document it. 

@rmackay9 - is this a known issue with Copter, and can we do anything about it?

atomictom:
Ah, sorry about that. I renamed `yaw()` to `condition_yaw()` just before posting it to be consistent with the name in the docs. [Here is an updated version](https://gist.github.com/atomictom/b10fd1af66443ddbe1fb) for anyone who is interested.

atomictom:
Also, I'm sorry, I did not mean to close it...

eliao:
I believe @asetapen encountered this yesterday and has some additional details.  Apparently this only happens after using a guided takeoff.

hamishwillee:
This will be resolved by https://github.com/dronekit/dronekit-python/pull/307

asetapen:
[...]

PR #307, opened 2015-09-08
Adds note to clarify point in #285

hamishwillee:
Clarifies that yaw cannot be set in guided mode until the first movement command has been received.


tcr3dr:
Merged into master.

PR #327, opened 2015-09-30
Assign a dedicated, public version.

tcr3dr:
This will supercede the other git tag.


hamishwillee:
Awesome. Should the module name also change to dronekit from droneapi at this point?

tcr3dr:
@hamishwillee Actually you're right, that should go in this commit as well.

hamishwillee:
Excellent!

PR #332, opened 2015-10-09
Adds ekf_of flag.

tcr3dr:


tcr3dr:
Review? @mrpollo

dronekit/__init__.py
```diff
@@ -232,6 +234,13 @@ def set(chnum, v):
             self.rngfnd_distance = m.distance
             self.rngfnd_voltage = m.voltage
             self.__on_change('rangefinder')
+        elif typ == "EKF_STATUS_REPORT":
+            # use same check that ArduCopter::system.pde::position_ok() is using
+            if self.status.armed:
+                self.ekf_ok = ((m.flags&ardupilotmega.EKF_POS_HORIZ_ABS) > 0) and (m.flags&ardupilotmega.EKF_CONST_POS_MODE == 0)
+            else:
+                self.ekf_ok = ((m.flags&ardupilotmega.EKF_POS_HORIZ_ABS) > 0) or ((m.flags&ardupilotmega.EKF_PRED_POS_HORIZ_ABS) > 0)
```
mrpollo: how can we make this block more legible?

dronekit/__init__.py
```diff
@@ -110,6 +111,7 @@ def __init__(self, master, status_printer=None):
         self.epv = None
         self.satellites_visible = None
         self.fix_type = None  # FIXME support multiple GPSs per vehicle - possibly by using componentId
+        self.ekf_ok = False
```
mrpollo: what do you think of making an object or class for `ekf` and exposing `status` attribute, we might want to add more data to this object later

tcr3dr: Opened another issue for that. (will want to keep ekf_ok for the betas)

https://github.com/dronekit/dronekit-python/issues/336

hamishwillee:
@tcr3dr 

1. This is inconsistent in that it returns False rather than None if we haven't yet received a message
1. I assume you want this documented? If so, would I be correct in thinking that this creates a boolean `Vehicle.ekf_ok` attribute that is True whenever EKF is reporting position is OK (ie within acceptable variance)?
  * Is it worth waiting until end of betas to document?
1. Test code in place?

tcr3dr:
@hamishwillee Actually not documenting this is fine; this is for compat with Solo's private extensions so I wanted to keep that code in place.

However, let's carry over these concerns to https://github.com/dronekit/dronekit-python/issues/336 for an actual design.

hamishwillee:
Sounds good! Thanks.

PR #333, opened 2015-10-09
Local position

squilter:
Listens for `Local_POSITION_NED` messages and parses them into a new local_position class.

### DIFF
--- a/tests/sitl/test_localposition.py
+++ b/tests/sitl/test_localposition.py
@@ -1,0 +1,30 @@
+from dronekit import connect
+from dronekit.tools import with_sitl
+import time
+from nose.tools import assert_not_equals
+
+@with_sitl
+def test_timeout(connpath):
+    v = connect(connpath, await_params=True)
+
+    # print "Basic pre-arm checks"
+    # Don't let the user try to fly autopilot is booting
+    if v.mode.name == "INITIALISING":
+        # print "Waiting for vehicle to initialise"
+        time.sleep(1)
+    while v.gps_0.fix_type < 2:
+        # print "Waiting for GPS...:", vehicle.gps_0.fix_type
+        time.sleep(1)
+
+    v.armed = True
+    v.flush()
+    
+    while not v.armed:
+        time.sleep(1)
+
+    time.sleep(1)
+    
+    #north, east, and down are initialized to None.  Any other value suggests that a LOCAL_POSITION_NED was received and parsed.
+    assert_not_equals(v.location_local.north, None)
+    assert_not_equals(v.location_local.east, None)
+    assert_not_equals(v.location_local.down, None)
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -80,6 +80,10 @@ class MPFakeState:
         self.lon = None
         self.alt = None
 
+        self.north = None
+        self.east = None
+        self.down = None
+
         self.vx = None
         self.vy = None
         self.vz = None
@@ -175,6 +179,11 @@ class MPFakeState:
             (self.lat, self.lon) = (m.lat / 1.0e7, m.lon / 1.0e7)
             (self.vx, self.vy, self.vz) = (m.vx / 100.0, m.vy / 100.0, m.vz / 100.0)
             self.__on_change('location', 'velocity')
+        elif typ == 'LOCAL_POSITION_NED':
+            self.north = m.x
+            self.east = m.y
+            self.down = m.z
+            self.__on_change('local_position')
         elif typ == 'GPS_RAW':
             pass # better to just use global position int
             # (self.lat, self.lon) = (m.lat, m.lon)
--- a/dronekit/__main__.py
+++ b/dronekit/__main__.py
@@ -20,7 +20,8 @@ def demo(local_connect):
 
     # Get all vehicle attributes (state)
     print "\nGet all vehicle attribute values:"
-    print " Location: %s" % v.location
+    print " Global Location: %s" % v.location_global
+    print " Local Location: %s" % v.location_local
     print " Attitude: %s" % v.attitude
     print " Velocity: %s" % v.velocity
     print " GPS: %s" % v.gps_0
--- a/dronekit/lib/__init__.py
+++ b/dronekit/lib/__init__.py
@@ -117,9 +117,9 @@ class Attitude(object):
     def __str__(self):
         return "Attitude:pitch=%s,yaw=%s,roll=%s" % (self.pitch, self.yaw, self.roll)
 
-class Location(object):
+class LocationGlobal(object):
     """
-    A location object.
+    A global location object.
 
     The latitude and longitude are relative to the `WGS84 coordinate system <http://en.wikipedia.org/wiki/World_Geodetic_System>`_.
     The altitude is relative to either the *home position* or "mean sea-level", depending on the value of the ``is_relative``.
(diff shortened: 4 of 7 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 .project
 .pydevproject
 dronekit.egg-info
+droneapi.egg-info
```
tcr3dr: This isn't needed.

tcr3dr:
Looks good! You can ignore Circle CI failing, we'll rerun it with credentials before merging.

Can you add a SITL test to go along with this?

And bikeshedding: Do we want `LocalPosition.north, .down, .east`, instead of `.N` `.E` `.D`?

mrpollo:
++ to `north`, `down`, `east` instead of `N`, `E`, `D`

squilter:
Yep, tested with SITL already   I can rename n,e,d and remove the first commit later today.

tcr3dr:
@squilter Oh, I mean an automated `assert`-based test like https://github.com/dronekit/dronekit-python/blob/master/tests/sitl/test_12.py. Anything file in that folder gets tested by CI.

mrpollo:
something simple like this would work (assuming `assertIsInstance` exists)

```python
@with_sitl
def test_local_position(connpath):
  v = connect(connpath, await_params=True)

  assertIsInstance(v.local_position, LocalPosition)
```

**Note:** this is just an example

squilter:
Also, should north, east and down be initialized to 0 instead of None?  The vehicle doesn't start sending position until it is armed (unlike global positions).

squilter:
What is up with the test numbers?  Am I supposed to make a new test file, and if so, what number do I give it?

tcr3dr:
By default we seem to use None, so let's stick with that.

The test numbers refer to specific bugs. You can give a title like `test_localposition.py` though.

squilter:
got it. one sec.

tests/sitl/test_localposition.py
```diff
@@ -0,0 +1,22 @@
...
+from nose.tools import assert_not_equals
+
+@with_sitl
+def test_timeout(connpath):
+    v = connect(connpath, await_params=True)
+
+    v.armed = True
+    v.flush()
+
+    time.sleep(5)
```
mrpollo: @tcr3dr 5 more secs on our tests, have you played with this number before to try to tune it?

squilter: I have not tried to tune it.  But I know that the boot sequence changes a lot.  Sitl used to calibrate the compass on boot for like 3 seconds.  Now it's faster.  But now the EKF needs to align on boot (which takes a while).  I used 5 seconds because sitl might take longer tomorrow than it does today, and that will be an annoying thing to debug.

squilter:
Test failed.  What's up with `Link timeout, no heartbeat in last 5 seconds`

tcr3dr:
@squilter The Appveyor Windows environment really hates SITL. Re-running it to see if that fixes it.

hamishwillee:
I'm slightly surprised we did it this way rather than fixing Vehicle.location to take a frame. I guess this is more flexible.

squilter:
Local position comes in a separate message.  So if we set up a callback for a new `location`, then the callback will get (unnecessarily) called 5 times per second.  That's why I thought to make a new object.

squilter:
Actually that's not a good reason.  @tcr3dr Should I just put north, east and down into the Location object?  I can just create a separate callback called local_position without creating a different object for it.

hamishwillee:
We've kind of got three choices

1. One location object that we set the "type" using a frame setter (ie "set we want NED" or infer frame from current movement mode). Location returns its frame.
1. A location object with multiple types: Vehicle.location.local and Vehicle.location.global and Vehicle.location.terrain.
1. Separate location objects as currently implemented, but in this case I'd still be tempted to rename Vehicle.location to Vehicle.location_global or similar

Philisophically I prefer "1" but I think it will be much harder to implement and fact is, there will be cases where the same computer might want to work in both global and NED frames. I think the "right" think to do would be "2" but this is a break and requires some changes all over the place. 

@tcr3dr @mrpollo Thoughts?

squilter:
I don't like that 2 and 3 will break support.  With 1, what if a user wants relative positions and global positions?

squilter:
@tcr3dr My test finally passed.  The goto test is failing for some reason.  

@hamishwillee I like the idea of adding North, East and Down as new field to the Location class.  It wouldn't break anything.

hamishwillee:
I much prefer "2" - changing Vehicle.location to Vehicle.location.global (.lat .lon .alt .isRelative - or frame) and Vehicle.location.local (.north .east .down). Yes it is an API break, but a) a major version change is where you make such breaks, b) it is a clean and "extensible" design. 

Adding north, east, down to Location isn't "bad" either but seems like messy API design to me - and perhaps not very good if we add support for location in other frames.

tcr3dr:
@squilter I'll take care of the failing tests; I'll want to rerun them on a true branch to pass Circle CI anyway.

squilter:
@tcr3dr @hamishwillee So should I move N,E,D to somewhere else?

hamishwillee:
@squilter Leave alone for the moment - @mrpollo or @tcr3dr need to comment on which option.

tcr3dr:
Okay, I've forked master to a separate https://github.com/dronekit/dronekit-python/tree/dkpy2-compat branch. This will track the -beta versions on pypi that aim for Solo compatibility, but we should be ready to consider backwards-incompatible changes on master now.

I am fine with `vehicle.location.global` and `vehicle.location.local`, I think this makes sense and I would love to land this change now for 2.0.0 rather than have two slightly asymmetrical ways of dealing with position. @mrpollo ?

mrpollo:
Let's do it, `vehicle.location.global` and `vehicle.location.local` both sound good to me

squilter:
Ok, I'll push to this branch in a few minutes

squilter:
`global` is a reserved keyword in python.  Ideas?

tcr3dr:
Eugh. Should we settle for `location_local` and `location_global` ?

squilter:
seems weird to type `v.location.location_local`.  But I haven't thought of anything better yet.

mrpollo:
`v.location.get_global`?

squilter:
`v.location.world` or `v.location.earth`?

tcr3dr:
Oh, I meant `vehicle.location_local` and `vehicle.location_global`.

squilter:
oh, that works.

hamishwillee:
A bit of a cheat, but how about ``vehicle.location.global_gps   [.lat| lon | alt | isRelative]``?

I slightly prefer this but can live with ``vehicle.location_local`` and ``vehicle.location_global``

@tcr3dr -final word?

squilter:
@tcr3dr Can you restart appveyor? it seems to have messed up again

hamishwillee:
@squilter We decided on the pattern like vehicle.location.local_frame, vehicle.location.global_frame etc.  I believe that @tcr3dr is planning to take this forward because he's doing some rapid iteration of the internal implementation that will otherwise break this and waste your time.

squilter:
Okay. I had it working, so it shouldn't be too hard for me to get it working again if you want. Just let me know if/when you want it.

tcr3dr:
Reopened as https://github.com/dronekit/dronekit-python/pull/351 (whoop, just about to get around to this!)

PR #330, opened 2015-10-08
Implements waypoints in DKPY2.

tcr3dr:
This also removes some deprecated properties (RC, __param), redirecting them to the module object.

Issue #329, opened 2015-10-08
DKPY2: Refactor RC override examples


tcr3dr:
I think it's appropriate to remove the RC override examples in DKPY2.

My primary motivation is not that this is unsafe user behavior (it can be but that's not relevant), but rather that this does not work on Solo, which has its own pipeline for RC controls. As a result, we don't have a generic way of this working even in the case of Python code, so we may want to approach this differently in the future.

For example, the `dronekit.sitl` package can include its own RC functions (the MAVProxy module actually is fixed to SITL output anyway); `dronekit.solo` could do likewise. But Dronekit core would not include these.

tcr3dr:
Bonus: I don't have to re-implement the RC module, which is currently unimplemented.

hamishwillee:
@tcr3dr @mrpollo I disagree. This will work on most platforms and there is a standard way to do it. One device should not fragment the core so that you have to have an extra library for every variant.

Is it possible to define an API that would work for Solo and "generic MAVLink"?

If you don't want to implement this module for v2.0 could we still have the API definition and Vehicle.channel_override can return 'None". I will then document it as not supported in the current version.

hamishwillee:
or I can implement it ... if the module "pattern" you're using is obvious so I can copy it.

hamishwillee:
@tcr3dr Counter-argument? IMO this is something used by developers now for "good reasons". If its not going to work on Solo that is a pain, but we shouldn't hobble DKPY2.0 for other vehicles.

eliao:
I think the APM devs would heavily discourage any usage of RC Override.

hamishwillee:
@eliao They absolutely do encourage **mis-use**. The main use case is for the creation of virtual sticks, and you could reasonably argue that on a platform that is intended for Companion Computers it isn't needed. But ... some people do want to use them and there have been a couple of cases where I couldn't actually find an alternative way of doing something without them.

Chambana:
Regardless of whether the examples are pulled from the documentation, please leave the RC override capability in API for development purposes.

tcr3dr:
Agreed that these need to exist for development purposes and should not be removed.

hamishwillee:
#335 adds the rc-override code/docs back as a "standalone" example. It isn't tested yet but this can probably still be merged.

tcr3dr:
See #427.

hamishwillee:
Good. This was done. Will have to be done again for #427.

PR #335, opened 2015-10-13
Adds stand-alone channel overrides example

hamishwillee:
This adds a standalone channel overrides example and example document to address #329. This completes the job of #330 by removing the channel override information from the vehicle state example and guide pages. It also tidies up the "running the example" section on the vehicle state example.

Note that the new example and example page has a big disclaimer saying that this shouldn't be used, but that if you need to you should post an issue explaining why so we can find a better alternative. I don't have a separate guide section on this because it isn't something we want to promote.

This has not yet been tested. The example page(s) "guess" at what the output should look like. I will test and update those sections as soon as there is a DK version I can use.

### DIFF
--- a/docs/examples/channel_overrides.rst
+++ b/docs/examples/channel_overrides.rst
@@ -32,7 +32,7 @@ In summary, after cloning the repository:
 
    .. code-block:: bash
 
-       cd dronekit-python\examples\channel_overrides\
+       cd dronekit-python/examples/channel_overrides/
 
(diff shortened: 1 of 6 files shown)


hamishwillee:
This also now tidies up the vehicle state example and docs to remove the MAVLInk message stuff that is no longer needed now that is in its own example. @mrpollo @tcr3dr Can I get this merged please?

PR #340, opened 2015-10-13
Abstracts MAVLink listening into observers, enabling dronekit-python-solo extension.

tcr3dr:
See the test in tests/sitl/test_vehicleclass.py for an example, also https://github.com/dronekit/dronekit-python-solo/blob/master/dronekit_solo/__init__.py


hamishwillee:
Very pretty. A bit of Python magic, but easy to copy/paste/extend.

PR #344, opened 2015-10-14
Adds Solo-specific system_status impl.

tcr3dr:
This is not meant to be a final implementation of system_status, but a stub one that allows us to achieve parity with Solo's private DK fork. We can continue to refine this design in the later RC's.

This will precede 2.0.0b4, and will probably be the last beta on master.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -193,8 +193,11 @@ class MPFakeState:
             self.level = m.battery_remaining
             self._notify_attribute_listeners('battery')
 
+        self.system_status = None
+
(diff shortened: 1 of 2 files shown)


mrpollo:
lgtm, just FYI not really solo specific, this is a patch to sync with Solo dk-py but this messages are generic to Ardupilot

tcr3dr:
Yeah, Solo-specific in that the only existing impl is for that (and not yet abstracted or nothin')

Future travelers: This doesn't need to be documented until it's implemented properly.

PR #347, opened 2015-10-14
Adds interfaces for Dronekit-SITL

tcr3dr:
*Goal:* Enable pysim support for Copter 3.2 and Solo builds in DroneKit-SITL. https://github.com/dronekit/dronekit-sitl/pull/47

This adds

```
vehicle.parameters.set(name, value, retries=*)
```

This allows the number of retries to be specified. 0 is fire and forget. Normal is 3, i.e. wait three intervals up to three seconds each to see if it is loaded. This exposes the underlying machinery for parameter setting via `vehicle.parameters[name] = value`.

This also adds

```
vehicle.close()
```

Whose behavior is approximately correct, though should probably go through a round of bikeshedding to ensure it's correct.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -548,6 +548,7 @@ class MPFakeState:
         t = Thread(target=mavlink_thread)
         t.daemon = True
         t.start()
+        self.mavlink_thread = t
 
         # Wait for first heartbeat.
         while True:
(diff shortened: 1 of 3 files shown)


mrpollo:
awesome work

:+1: :cake:

hamishwillee:
@tcr3dr Very nice. Would be handy to get something like this for armed and mode setting.I like returning an error over raising an exception.

I added label `[needs] docs`. 

Can you explain the Vehicle.close semantics - when/why should this be called, and what exactly is the state of the vehicle after doing so? Is this called automatically when a script exits/falls through? etc.

PR #349, opened 2015-10-15
Adds rate param to connect.

tcr3dr:
Needed by Donkey Kong, possibly up for formalization as a DK feature.


hamishwillee:
Hope so. No docs at this time then?

PR #359, opened 2015-10-21
Add vehicle.close to vehicle_state example

hamishwillee:
Minor fix the the vehicle_state.py example to ensure vehicle.close() is called before exiting.


hamishwillee:
Close this. The work was done in #335

Issue #355, opened 2015-10-18
DKPY2 Can't clear waypoints


hamishwillee:
The following test code returns and error that `AttributeError: MPFakeState instance has no attribute 'vehicle'`  when I try to access Vehicle.commands.clear. Actually downloading of commands and reading home location works fine.

Test code:

```
from dronekit import connect
import time
import math
from dronekit.lib import VehicleMode, Location, Command
from pymavlink import mavutil


#Set up option parsing to get connection string
import argparse  
parser = argparse.ArgumentParser(description='Print out vehicle state information. Connects to SITL on local PC by default.')
parser.add_argument('--connect', default='127.0.0.1:14550',
                   help="vehicle connection target. Default '127.0.0.1:14550'")
args = parser.parse_args()

# Connect to the Vehicle
print 'Connecting to vehicle on: %s' % args.connect
vehicle = connect(args.connect, await_params=True)

cmds = vehicle.commands
cmds.download()
cmds.wait_valid()

print "About to test clear"
vehicle.commands.clear()
vehicle.flush()
print "Clear succeeded"
```

Output of test code
```
About to test clear
Traceback (most recent call last):
  File "E:\deleteme\dronekit-python\examples\mission_basic\test_mission_clear.py", line 29, in <module>
    vehicle.commands.clear()
  File "build\bdist.win-amd64\egg\dronekit\module\api.py", line 95, in clear
AttributeError: MPFakeState instance has no attribute 'vehicle'
```

This causes the **mission_basic.py** and **flight_replay.py** examples to fail in dkpy2.0

hamishwillee:
@tcr3dr When you have time, I've marked this as urgent because it breaks two examples.

tcr3dr:
See https://github.com/dronekit/dronekit-python/pull/374 for the fix.

hamishwillee:
@tcr3dr Thanks - agreed fixed. Unfortunately now we can't upload waypoints - see #376

PR #360, opened 2015-10-21
Fix up flight_replay.py doc/example to DKPY2. Note, still has bug #355

hamishwillee:
This fixes up the flight_replay example and associated docs to be "DKPY2 compliant". This includes adding vehicle.close and connect, and making the example docs follow the same pattern as the other example docs.

NOTE:

* This does not work! There is an exception due to #355 . The docs have a note to this effect. When the bug is fixed we'll have to revisit and retest the docs/code.
* This hamishwillee dronecode key and api. The old version failed because droneshare now insists on a key with the correct format.
* Even though this does not "work", it is a better example to have for DKPY2 for people to learn from. The old example would not work either due to the bad key!

### DIFF
--- a/docs/examples/flight_replay.rst
+++ b/docs/examples/flight_replay.rst
@@ -35,7 +35,7 @@ In summary, after cloning the repository:
 
    .. code-block:: bash
 
-       cd dronekit-python\examples\flight_replay\
+       cd dronekit-python/examples/flight_replay/
 
 
 #. Start the example, passing the :ref:`connection string <get_started_connect_string>` 
@@ -63,8 +63,8 @@ On the command prompt you should see (something like):
 
     \dronekit-python\examples\flight_replay>flight_replay.py
     Connecting to vehicle on: 127.0.0.1:14550
-    >>> ☺APM:Copter V3.4-dev (e0810c2e)
-    >>> ☺Frame: QUAD
+    >>> APM:Copter V3.4-dev (e0810c2e)
+    >>> Frame: QUAD
     JSON downloaded...
     Generating 95 waypoints from replay...
     ...
--- a/docs/examples/flight_replay.rst
+++ b/docs/examples/flight_replay.rst
@@ -2,57 +2,95 @@
 Example: Flight Replay
 =========================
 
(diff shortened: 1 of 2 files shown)


docs/examples/flight_replay.rst
```diff
@@ -2,57 +2,95 @@
...
+    explains how to set up *Mission Planner* to view a vehicle running on the simulator (SITL).
+
+On the command prompt you should see (something like):
+
+.. code-block:: bash
+
+
+    \dronekit-python\examples\flight_replay>flight_replay.py
+    Connecting to vehicle on: 127.0.0.1:14550
+    >>> ☺APM:Copter V3.4-dev (e0810c2e)
```
tcr3dr: Can you remove these smiley faces? They're just a bug on master :) Was an errant, random character.

examples/flight_replay/flight_replay.py
```diff
@@ -63,7 +71,11 @@ def download_messages(mission_id, max_freq = 1.0):
 def replay_mission(payload):
     """Given mission JSON, set a series of wpts approximating the previous flight"""
     # Pull out just the global position msgs
-    messages = payload['messages']
+    try:
+        messages = payload['messages']
+    except:
+        print "Exception: payload from site is: %s" % payload
+        exit()
```
tcr3dr: Should this be `sys.exit`?

examples/flight_replay/flight_replay.py
```diff
@@ -94,17 +106,20 @@ def replay_mission(payload):
...
+max_freq = 0.1
+json = download_messages(mission_id, max_freq)
+print "JSON downloaded..."
+replay_mission(json)
+
+
+#Close vehicle object before exiting script
+print "Close vehicle object"
+vehicle.close()
+print("Completed")
```
tcr3dr: No parens.

examples/flight_replay/flight_replay.py
hamishwillee: Yes. Wonder why exit() does actually work?

hamishwillee:
Fixed. Ready to merge.

Issue #357, opened 2015-10-20
Mode not changed when message sent inside drone delivery example


hamishwillee:
@tcr3dr 

I can't get the drone_delivery example to work. I've [updated it in this gist](https://gist.github.com/hamishwillee/22102fe31c3473c3a36e). Unfortunately it won't take off because the command to change the mode to guided never succeeds - it just sits there in the wrong mode. 

The same code run in a "normal" environment does work - you can test this by just enabling [this line](https://gist.github.com/hamishwillee/22102fe31c3473c3a36e#file-drone_delivery-py-L82).

I have also seen cases while experimenting where I changed the mode manually in the simulator via mavproxy, and this change was not detected - though the system sat there telling me it was getting the wrong mode.

I've tried lots of "trial and error" but I can't work out why it works in one case and not the other. Needs debugging from someone who can check the messages are actually being sent.

tcr3dr:
Addressed by #361.

PR #361, opened 2015-10-21
Minimal changes to drone_delivery for dkpy2

hamishwillee:
This updates the example to use the new `connect()` and close methods and fixes a few obvious bugs/inconsistencies. The  example does not take off due to reasons given in #357 (this is added as a warning in the docs).

I also updated the docs to use our "standard" format and put in additional information about how to install the dependencies (tested on Windows).

Obviously this will require further updates when #357 is fixed, but is better than leaving "as is".

### DIFF
--- a/docs/examples/drone_delivery.rst
+++ b/docs/examples/drone_delivery.rst
@@ -2,7 +2,14 @@
 Example: Drone Delivery
 ===========================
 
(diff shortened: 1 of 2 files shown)


mrpollo:
we should update the docs @hamishwillee http://python.dronekit.io/examples/drone_delivery.html

PGSK1234:
I used the example python for Drone_Delivery from the Master repository and saw the same behavior with the mode not moving beyond "GUIDED". I am new to Github, Python... I tried using the attached python file which i thought had the fix... but see this error.. Please help. 

image
Error
[drone_delivery_test.txt](https://github.com/dronekit/dronekit-python/files/6836691/drone_delivery_test.txt)

PR #377, opened 2015-10-27
Removes "MP" classes

tcr3dr:
This PR removes all the "MP" class duplication and merges the classes into actual definitions.

It also removes the "API" methods that are unused.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -5,9 +5,12 @@ import sys
 import os
 import platform
 import re
+import dronekit.module.api
 from pymavlink import mavutil, mavwp
 from Queue import Empty
 from pymavlink.dialects.v10 import ardupilotmega
+from Queue import Queue
+from threading import Thread
 import types
 
 if platform.system() == 'Windows':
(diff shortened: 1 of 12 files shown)


hamishwillee:
@tcr3dr Hallelujah! 

How will people import dronekit following this change? import dronekit? or import dronekit.lib

Where will the docs/definitions live - currently everything in dronekit.lib \_\_init\_\_.py but because not everything was in the "interface" I've had to invent some documentation (upshot, best if we have a proper stucture where the init.py really does define an interface.

tcr3dr:
@hamishwillee I want to isolate this PR from the .lib. removal, but yes, dronekit.Vehicle and dronekit.VehicleMode should be the new idiom.

I need to root around some other Python libraries for how their structure does it, but yes, we should look at breaking apart our massive __init__.py file into several smaller files at the next availability.

hamishwillee:
@tcr3dr Sounds great. As long as you update our test code to match  the idioms and it has enough coverage ... I can work out what is needed and extend changes to docs

PR #380, opened 2015-10-28
Improve about section of docs 

hamishwillee:
This updates the About section of docs for DKPY2. 

1. Tidies the release notes a little. 
1. Updates the overview. 

The major changes were to:

* make it clear that DK is not really compatible with every single MAVLink device. I did this by removing words like "all" and adding a note indicating what we validate against.
* make it clear that we no longer get back a "list of devices" when connecting, and that we can connect to multiple devices from script
* Remove/reduce the "air interface" branding.

### DIFF
--- a/docs/about/github_latest_release.txt
+++ b/docs/about/github_latest_release.txt
@@ -17,7 +17,7 @@ Source Code
 -----------
 
 
-Source code is available at <https://github.com/dronekit/dronekit-python/releases/tag/v1.5.0>.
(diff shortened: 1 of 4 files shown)


hamishwillee:
@tcr3dr This can be merged. The section does need some work still - basically we may need to change the support link to point to forums. You will also need to update release notes, but that can be done as a separate pull.

mrpollo:
lgtm

PR #207, opened 2015-07-16
[WIP] Adds separate .home_location from .commands array. Fixes #132

tcr3dr:
Work in progress tests and fixes.

PR #384, opened 2015-10-29
Moves home_location out of commands array.

tcr3dr:
Replaces #207. See #132.


tcr3dr:
One issue is that `home_location` is None until `commands.download()` is called. hm.

hamishwillee:
@tcr3dr Very cool. 

> One issue is that home_location is None until commands.download() is called. hm.

You could kick off a commands download every time this is accessed "within a timeframe"? A bit ugly, particularly if there are a lot of points - since waiting for download is blocking. I'd be tempted to just leave as is and solve with documentation.

What we should do though is make this attribute writeable. I am *fairly* sure that the command to set the home position is in Copter/Rover/Plane already.

Something to think about ...

Support for requesting the HOME_POSITION using MAV_CMD_GET_HOME_POSITION [is in master](https://github.com/diydrones/ardupilot/blob/master/ArduCopter/GCS_Mavlink.cpp#L1389) but not in any release yet. Obviously using this is the "rght" thing to do, but will not be backward compatible with current releases. I guess you have a clever plan to deal with that.

hamishwillee:
@tcr3dr This build crashes if clear() is called on an empty object. So basically the below crashes - but it should work - you shouldn't need to do anything special to get a vehicle.commands object ready to clear (uncommenting the lines works)

```
cmds = vehicle.commands
#cmds.download()
#cmds.wait_valid()
cmds.clear()
vehicle.flush()    
```

The error shown is:
```
  File "E:\deleteme\dronekit-python\examples\mission_import_export\test.py", line 22, in <module>
    cmds.clear()
  File "build\bdist.win-amd64\egg\dronekit\lib\__init__.py", line 1057, in clear
  File "C:\Apps\WinPython-64bit-2.7.10.2\python-2.7.10.amd64\lib\site-packages\pymavlink-1.1.62-py2.7.egg\pymavlink\mavwp.py", line 60, in add
    w.comment = comment
AttributeError: 'NoneType' object has no attribute 'comment'
```
Copyable test code below.
```
from dronekit import connect
from dronekit.lib import Command
from pymavlink import mavutil


#Set up option parsing to get connection string
import argparse  
parser = argparse.ArgumentParser(description='Example that demonstrates mission import/export from a file. Connects to SITL on local PC by default.')
parser.add_argument('--connect', default='127.0.0.1:14550',
                   help="vehicle connection target. Default '127.0.0.1:14550'")
args = parser.parse_args()

# Connect to the Vehicle
print 'Connecting to vehicle on: %s' % args.connect
vehicle = connect(args.connect, await_params=True)
    
print 'Clear existing mission'
cmds = vehicle.commands
#cmds.download()
#cmds.wait_valid()
cmds.clear()
vehicle.flush()    

        
#Close vehicle object before exiting script
print "Close vehicle object"
vehicle.close()
```

hamishwillee:
@tcr3dr ``vehicle.home_location`` should be of type ``LocationGlobal`` - not a MISSION_ITEM :-)

hamishwillee:
@tcr3dr ``Vehicle.commands`` needs to be iterable and sliceable so that it can be manipulated/changed. In the **mission_import_export.py** we use a slice to get rid of the home location (arguably we shouldn't need to do this any more, but it is still useful to be able to manipulate data sets). We iterate the rest in order to add the messages to a list, which we can then manipulate.

I ran **mission_import_export.py** and got ``TypeError: unsupported operand type(s) for -: 'slice' and 'int'``. Thinking I no longer need that I iterate over the whole list (see line after  commented slice on cmds[1:] below)

```
def download_mission():
    """
    Downloads the current mission and returns it in a list.
    It is used in save_mission() to get the file information to save.
    """
    missionlist=[]
    cmds = vehicle.commands
    cmds.download()
    cmds.wait_valid()
#    for cmd in cmds[1:]:  #skip first item as it is home waypoint.
    for cmd in cmds:  #iterate over all waypoints now the first is no longer "home"
        print cmd   
        missionlist.append(cmd)
    return missionlist
```
The result was an endless loop after the end of the cmds list - see below. The first waypoint still appears to be the home location?
```
Save mission from Vehicle to file: exportedmission.txt
MISSION_ITEM {target_system : 255, target_component : 0, seq : 8, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : 0.0, y : 0.0, z : 0.0}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 0, frame : 0, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3632621765, y : 149.165237427, z : 583.989990234}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 1, frame : 0, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3605003357, y : 149.172363281, z : 747.0}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 2, frame : 3, command : 22, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3598327637, y : 149.16633606, z : 100.0}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 3, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3634872437, y : 149.167205811, z : 100.0}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 4, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3554916382, y : 149.16960144, z : 100.0}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 5, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3550720215, y : 149.175842285, z : 100.0}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 6, frame : 3, command : 113, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : 0.0, y : 0.0, z : 22222.0}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 7, frame : 0, command : 115, current : 0, autocontinue : 1, param1 : 2.0, param2 : 22.0, param3 : 1.0, param4 : 3.0, x : 0.0, y : 0.0, z : 0.0}
MISSION_ITEM {target_system : 255, target_component : 0, seq : 8, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : 0.0, y : 0.0, z : 0.0}
None
None
None
None
None
... (forever)
```

hamishwillee:
@tcr3dr excellent! I'll try test this today.

BTW, if you clear then add a command, without flushing between, will the autopilot successfully clear then add the commands? If so, then no change needed. If there is potential for race condition then it seems to me that you will always need to flush and possibly delay between commands, and arguably we should then always call flush inside clear so that users don't need to.

hamishwillee:
@tcr3dr Thanks!

The good:

1. ``home_location`` is a ``LocationGlobal``, 
1. slicing and iterating work (great) 
1. if you iterate on an empty ``vehicle.commands`` object it does not crash :-)
1. You can now call "clear" on an empty vehicle.commands object and it "kind of" works (see below). - ie no need to call ``cmds.download()`` then ``cmds.wait_valid()`` just to clear the cmds.

The bad:

* The ``vehicle.commands`` array still contains the home location - is that intentional?
* If you clear the commands array it still has home location and an additional "rubbish" point which is sometimes the home location and sometimes" something else"
* If you download vehicle.commands you get two extra points, and the sequence number is wrong.

It may be that you still expect home location to be in the commands array. I hope not, but even so something is not quite right.

Some debug:

1. If you clear the commands then iterate you get two points. These are both for the home location - previously we had one value for the home location, and I had hoped that we'd get NONE!

   ```
DOWNLOAD CMD: MISSION_ITEM {target_system : 0, target_component : 0, seq : 0, frame : 0, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3632621765, y : 149.165237427, z : 583.989990234}
DOWNLOAD CMD: MISSION_ITEM {target_system : 0, target_component : 0, seq : 0, frame : 0, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3632621765, y : 149.165237427, z : 583.989990234}
```

   Note the sequence number =0

1. If you run the mission_import_export.py and iterate over the whole array downloaded you get this (space added between lines to shoe home location)
   ```
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 8, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : 0.0, y : 0.0, z : 0.0}
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 0, frame : 0, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3632621765, y : 149.165237427, z : 583.989990234}

DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 1, frame : 0, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3605003357, y : 149.172363281, z : 747.0}
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 2, frame : 3, command : 22, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3598327637, y : 149.16633606, z : 100.0}
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 3, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3634872437, y : 149.167205811, z : 100.0}
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 4, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3554916382, y : 149.16960144, z : 100.0}
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 5, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : -35.3550720215, y : 149.175842285, z : 100.0}
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 6, frame : 3, command : 113, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : 0.0, y : 0.0, z : 22222.0}
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 7, frame : 0, command : 115, current : 0, autocontinue : 1, param1 : 2.0, param2 : 22.0, param3 : 1.0, param4 : 3.0, x : 0.0, y : 0.0, z : 0.0}
DOWNLOAD CMD: MISSION_ITEM {target_system : 255, target_component : 0, seq : 8, frame : 3, command : 16, current : 0, autocontinue : 1, param1 : 0.0, param2 : 0.0, param3 : 0.0, param4 : 0.0, x : 0.0, y : 0.0, z : 0.0}
Close vehicle object
```

   The last 8 points are correct. The first two have stuffed up sequence numbers. The first point is empty/null point but the second is the home location (a bit odd). Again, I would expect neither of these to be output so users can be ignorant of the fact that the home location is stored as waypoint 0.

Does that make sense?

tcr3dr:
@hamishwillee Fixed the home_location errors.

I think that clear() then add() will not require a flush() inbetween. For example:

```
.clear()
.add(Command(...))
.flush()
```

This is the whole transaction, i.e. we don't talk to Pixhawk until the flush() command.

tcr3dr:
And for the hell of it, `vehicle.commands.upload` instead of `vehicle.flush`.

hamishwillee:
Looking good so far ...

1. If you get the homelocation before it has been set the value is lat=0.0,lon=0.0,alt=0.0,is_relative=False - is this intentional/should it be `None`?
1. the new upload function makes a lot more sense. Note, `flush` has not been removed
1. The sequence numbers returned are from 1 in command list (as you would expect). We might need to explain this in docs though.
1. There are some rounding "issues" in the round trip. If I get a file and upload -35.359831 then the value that comes back after downloading is -35.3598327637

hamishwillee:
@tcr3dr I added some documentation to go with this - updates to the auto mode docs, the main mission examples. Also added API ref for home_location. There is more to do, but feel free to merge what you have now if you feel the need. 

Also, see #390 which is the only bug I've found with this so far.

hamishwillee:
@tcr3dr Thats it, docs all up to date for ``vehicle.home_location`` and ``vehicle.commands`` changes in this PR (modulo the issue reports and comments here). It can all be integrated, ideally sooner rather than later!

tcr3dr:
@hamishwillee The rounding error is unavoidable. I believe all the params are encoded as 32-bit floats, not 64-bit doubles (Python's native datatype). This is demonstrated here:

```
>>> import struct
>>> struct.unpack("f", struct.pack("f", -35.359831))
(-35.359832763671875,)
```

A documentation-level advisory would help.

tcr3dr:
@hamishwillee 

> If you get the homelocation before it has been set the value is lat=0.0,lon=0.0,alt=0.0,is_relative=False - is this intentional/should it be None?

This is what's returned by the autopilot. Since the 0,0,0 tuple is "magical" (as far [as I can tell](https://github.com/diydrones/ardupilot/blob/b15bf3243ed085409f77d211c7795ec9bc0ad8cc/ArduCopter/commands_logic.cpp#L840)) I would be amenable to either returning None in this case (since that's our default) or returning what it does now.

tcr3dr:
@hamishwillee I'll wait for your review on this tonight that the two open bugs are resolved. Then I'm happy to merge.

/cc @mrpollo for additional code review.

tcr3dr:
@hamishwillee I don't want to remove `.flush()` for a bit (perhaps by 2.0.0 final) but we should definitely mark it as deprecated.

tcr3dr:
Branch fixed, again.

hamishwillee:
@tcr3dr 

>> If you get the homelocation before it has been set the value is lat=0.0,lon=0.0,alt=0.0,is_relative=False - is this intentional/should it be None?

>>This is what's returned by the autopilot. Since the 0,0,0 tuple is "magical" (as far as I can tell) I would be amenable to either returning None in this case (since that's our default) or returning what it does now.

I am not sure. My inclination is to leave this as is - on the assumption it might be useful to know that a home position has not yet been set.

hamishwillee:
@tcr3dr  

> The rounding error is unavoidable. I believe all the params are encoded as 32-bit floats, not 64-bit doubles (Python's native datatype). This is demonstrated here ...
  ...
  A documentation-level advisory would help.

Agreed, I will put a comment in the docs somewhere!

tcr3dr:
@hamishwillee This branch still has broken `.next` behavior but as a heads up, I will address it in a following PR; this one is getting unwieldly and I'd like to move fast with more refactoring.

hamishwillee:
Absolutely - looking forward to seeing this in a live branch. 

I presume that setting the home location is also still broken?

tcr3dr:
@hamishwillee Not sure. I believe not, but let me confirm on a new PR...

Issue #342, opened 2015-10-14
DKPY2: Await attributes and parameters


tcr3dr:
Bikeshedding time~~~!

I feel like we want a method `.await([types], defaults=True)` that does the following:

1. Waits for system_status to enter an initialized state. This is equivalent to a `.notify_change` event on `"initialized"`. (included by default when defaults=True)
1. Waits for parameters to be loaded. This is equivalent to a `.notify_change` event on `"parameters"`. (included by default when defaults=True)
1. Waits for what other defaults? (Armed? Mode? etc.)
1. Waits for an arbitrary list of `.notify_change` events. This encompasses attributes.

So an example vehicle state example may be:

```py
vehicle = connect('127.0.0.1:14550')
vehicle.await(['location', 'attitude', 'velocity', 'gps_0', 'groundspeed',
               'airspeed', 'mount_status', 'battery', 'rangefinder', 'mode', 'armed'])

print " Location: %s" % vehicle.location
print " Attitude: %s" % vehicle.attitude
print " Velocity: %s" % vehicle.velocity
print " GPS: %s" % vehicle.gps_0
print " Groundspeed: %s" % vehicle.groundspeed
print " Airspeed: %s" % vehicle.airspeed
print " Mount status: %s" % vehicle.mount_status
print " Battery: %s" % vehicle.battery
print " Rangefinder: %s" % vehicle.rangefinder
print " Rangefinder distance: %s" % vehicle.rangefinder.distance
print " Rangefinder voltage: %s" % vehicle.rangefinder.voltage
print " Mode: %s" % vehicle.mode.name    # settable
print " Armed: %s" % vehicle.armed    # settable
```

Wow, what a verbose function! But does it make sense? We want to *ensure* these values are populated. But this example then doesn't scale if we are always waiting for a rangefinder event that doesn't happen. Maybe we want to await a selection of always-on parameters, then time-await a selection of properties that may be optional?

```py
vehicle = connect('127.0.0.1:14550')
vehicle.await(['location', 'attitude', 'velocity', 'gps_0', 'groundspeed',
               'airspeed', 'mode', 'armed'])

print " Location: %s" % vehicle.location
print " Attitude: %s" % vehicle.attitude
print " Velocity: %s" % vehicle.velocity
print " GPS: %s" % vehicle.gps_0
print " Groundspeed: %s" % vehicle.groundspeed
print " Airspeed: %s" % vehicle.airspeed
[...]

PR #389, opened 2015-10-31
Changes parameter access behavior to not wait on key lookup.

tcr3dr:
This behavior seems fairly "magical", requires a hack in Solo, and also .wait_valid() in terms of parameters should be generalized in the `.wait('parameters')` magic (#342).

### DIFF
--- a/dronekit/lib/__init__.py
+++ b/dronekit/lib/__init__.py
@@ -855,11 +855,9 @@ class Parameters(HasObservers):
         self.__module = module
 
     def __getitem__(self, name):
-        self.wait_valid()
         return self.__module.mav_param[name]
 
     def __setitem__(self, name, value):
-        self.wait_valid()
         self.__module.param_set(name, value)
 
     def set(self, name, value, retries=3, await_valid=False):
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup, Extension
 import platform
 
-version = '2.0.0rc3'
+version = '2.0.0rc4'
 
 setup(name='dronekit',
       zip_safe=True,


hamishwillee:
@tcr3dr Was the ``wait_valid()`` waiting for ALL parameters to be downloaded or just the current parameter you queried? If the former then IMO it would be good to add #342 first so that we don't have to poll for the value.

More generally, the later case interests me. Yes we get all parameters on load, but how do we compensate for the possibility they might change - either due to a connected GCS or automatically in autotune?

tcr3dr:
@hamishwillee The old behavior of this was to wait for all parameters to be downloaded before `parameters['KEY']` returned anything.

So our choice is to:

1. Accept that accessing a parameter before it is downloaded will throw a KeyError rather than None (like other default attributes do), or
2. Accept that parameter access is not "cheap", i.e. can enter into a (potentially infinite!) spin-loop waiting for parameter download to occur

I'm strongly against the latter, since it makes property access magical and even network-dependent. It's better to explicitly say where the parameter download behavior should occur. Right now that is possible with the `await_params=True` flag, which we can update our examples to use.

Basically, I would want my script to a) connect to the drone, b) download the current state of all variables, and c) receive updates over time so that our state always mirrors that of the autopilot. I think that's the case right now with `connect(await_params)`, and is being extended more broadly in #342.

In summary, I would never call `.parameters[key]` without either explicitly waiting for the initial download, or accepting that we may not know of a parameter's existence yet and code accordingly. Solo, in fact, does the latter with a try: except: construct in its shots code(!).

hamishwillee:
@tcr3dr Thanks Tim - All examples already use ``await_params=True`` flag so if the "what we have now" isn't broken by this change it is good for me.

Essentially there are two models a developer might like. The one you describe after "Basically, I would ..." is one model, and the one everyone uses and understands at the moment. The other is where you allow connection without parameters, but people can get individual parameters when they need them. I think you're saying if that they try to get a parameter that has not yet downloaded this will fail.

Anyway, upshot at the moment is that I probably should document this expectation in http://python.dronekit.io/guide/vehicle_state_and_parameters.html#parameters - OK?

tcr3dr:
Gonna hold off on this pending more discussion.

hamishwillee:
Either way, we need to have a short discussion in the near future so you can confirm exactly what behaviour parameters has, so I can update the docs. 

Are we getting/do we need parameter observers? 

My gut feeling is that they are a "nice to have" not a "need" - you'd only really need an observer for maintaining a UI (ie for re-writing a GCS) and in most cases the UI for parameters would tend to be a static set of data that would be opened only for changes (so wouldn't need monitoring)/

tcr3dr:
Given that we have .set() and .get() now, it seems users have all the control they need to avoid auto-parameter download behavior. Going to close this.

hamishwillee:
Good idea

Issue #319, opened 2015-09-23
DKPY2 await_params=True should be the default


hamishwillee:
@tcr3dr Whatever await mechanism gets decided for awaiting, the default should be to ensure that if no await is specified ``connect()`` waits until all parameters are collected. That is consistent with what GCS like Mission Planner does, with DKPY1. Also it simplifies the documentation.

tcr3dr:
Going to close this in favor of #342.

PR #401, opened 2015-11-03
Extracts MPState object logic into Vehicle. Implements wait_ready.

tcr3dr:
This is a pretty heavy PR, but also the last major refactor I have planned.

Rather than obtusely split parameter logic between a MAVLink handler and a Vehicle object, the Vehicle object (through "message" handlers and "event loop" handlers) now implements all of its own core logic inside the class. There is still some slightly awkward coupling going on between the classes, but there's far less of it, and the code to manage properties is all in one place.

As a side effect of this, we're now able to track the status of most attributes, and as such have a working `wait_ready` implementation on the vehicle that serves as the basis for `await_param` (which should be deprecated and replaced) as well as other methods like `wait_valid`, etc.

Additionally, we get rid of the MAVLink rx handlers and only expose the `on_message` primitive, which acts as our MAVLink interaction story. Ideally, we can abstract the message contruction story as well, though I haven't put much thought into it.

### DIFF
--- a/dronekit/lib/__init__.py
+++ b/dronekit/lib/__init__.py
@@ -1,6 +1,6 @@
 # DroneAPI module
 
-import threading, time, math
+import threading, time, math, copy
 import CloudClient
 
 from pymavlink import mavutil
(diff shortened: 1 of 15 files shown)


tcr3dr:
See #319, #342 for more discussion.

hamishwillee:
Looks a lot simpler. What, if any testing/documentation do you "know" you need form me?

This merge misses a lot of the documentation changes we "temporarily lost" this morning - e.g I replaced use of flush() everywhere and had started to use home_location. I am assuming that will be fixed up?

dronekit/lib/__init__.py
```diff
@@ -273,28 +267,28 @@ def __init__(self):
...
         for watching several attributes.
-        
+
         The example below shows how to get callbacks for location changes:
 
         .. code:: python
 
             #Callback to print the location in global and local frames
-            def location_callback(location):
+            def location_callback(vehicle, location):
```
mrpollo: in this case this should be `def location_callback(vehicle, attr_name):` see `attr_name` instead of `location`, `attr_name` is a string

dronekit/lib/__init__.py
```diff
@@ -550,29 +526,355 @@ class Vehicle(HasObservers):
...
+            self._notify_message_listeners(msg.get_type(), msg)
+
+        self._lat = None
+        self._lon = None
+        self._vx = None
+        self._vy = None
+        self._vz = None
+
+        @self.message_listener('GLOBAL_POSITION_INT')
+        def listener(self, name, m):
```
mrpollo: i feel dirty by re-defining `def listener` should i call the doctor? or my psychiatrist?

mrpollo:
lots of changes, left some comments but non blocking on merge, this looks good :+1:

dronekit/lib/__init__.py
tcr3dr: You should call the Ghostbusters because names attached to python functions attached to decorators are not exposed to any scope!!! Spooky :ghost: :ghost: :ghost:

dronekit/lib/__init__.py
tcr3dr: NICE CATCH.

hamishwillee:
@tcr3dr Hey, but the docs still don't include my previous updates - e.g. 

```
        for cmd in cmds[1:]:  #skip first item as it is home waypoint.
             missionlist.append(cmd)		             missionlist.append(cmd)
```
The slice is no longer needed and was removed.

Do I need to go through all the examples and re-verify my changes went in?

tcr3dr:
@hamishwillee Hm. It may have been overwritten by git due to its proximity to a wait_valid invocation; my fault, though I can't find the original commit now where it was changed.

Can you find the commit in which it was modified? It should be this commit or one of it's parents, if I recall: https://github.com/dronekit/dronekit-python/commit/8958a38a5237be660ed48585d20c8f1c58a17e7a Or was it a different PR?

hamishwillee:
@tcr3dr The specific case I pointed to (cmds slicing) was fixed in https://github.com/dronekit/dronekit-python/commit/dd7058a5677fc85fc19f24d9e9aacd83899efc42 - but the changes in https://github.com/dronekit/dronekit-python/commit/8958a38a5237be660ed48585d20c8f1c58a17e7a need to be in there too (both of these were the problem commits that disappeared yesterday morning).

tcr3dr:
@hamishwillee I think the commits got pulled in. If you look in that commit, the `[1:]` slice was just accidentally left in: https://github.com/dronekit/dronekit-python/blob/dd7058a5677fc85fc19f24d9e9aacd83899efc42/docs/guide/auto_mode.rst

hamishwillee:
Excellent - sorry, I'll fix that!

PR #406, opened 2015-11-03
Supports Copter 3.3 and new DroneKit SITL.

tcr3dr:

PR #407, opened 2015-11-03
Reworks timeout logic. Fixes windows tests. Supports Copter 3.3 release in tests.

tcr3dr:
Motivated by the Windows tests often failing, I figured out what was breaking them (finally!) This patch does a few things:

* Switches our tests to dronekit-stil 3
* Switches tests to Copter-3.3 (not RC)
* Stops issuing as many 5s warnings
* Adds a `heartbeat_timeout` property to `connect()`, by default set to 30s (this is different from the 5s warning, in that it actually closes the connection)


tcr3dr:
This branch supercedes #406, since Linux was actually failing (somewhat consistently) on one of the old builds.

dronekit/test/sitl/test_110.py
```diff
@@ -7,18 +7,16 @@
 
 @with_sitl
 def test_110(connpath):
-    v = connect(connpath, wait_ready=True)
+    vehicle = connect(connpath, wait_ready=True)
```
mrpollo: aww yiss! no more `v`

hamishwillee:
Is this timeout just for connecting? If so, can user  set 0 to disable the timeout altogether (should they so wish)? 

If it is a general timeout, would we ever want to change this timeout after connecting?

tcr3dr:
@hamishwillee It can now be set to 0 (or None) to not trigger.

We will probably want the ability to change this timeout after connecting, but I want to expose that in another PR (or issue).

mikerob:
Looking forward to testing - do I just update Dronekit via pip to get the latest?

Also, how do I access the documentation as listed in the top of lib/_init_.py?  There's a lot of good info that doesn't seem to show up anywhere?

"Connect"description / doco in lib/_init_.py doesn't include timeout (line 46)

It seems that if the connection times out, this happens:

raise APIException('Timeout in initializing connection.')

The timeout.py sample doesn't use this, it seems to just end when / if it times out, which at least is better than hanging.  Is the intent that you use the raised exception to identify that the connect has failed and handle it that way, or is there a parameter within "Vehicle" that indicates it's not connected?

hamishwillee:
@mikerob It is unlikely that a new release has appeared for this on pip yet. YOu can however build dronekit from github source using the [instructions here](http://python.dronekit.io/contributing/developer_setup_windows.html) (on windows). You may have to uninstall dronekit first (``pip uninstall dronekit``).

The documentation hasn't been rebuilt yet (@tcr3dr - worth doing another docs build soon!). It has much improved, but does not yet include the timeout docs (this only went in 3 hours ago!). If you want to build and test the docs locally yourself it is quite easy - [see these instructions](http://python.dronekit.io/contributing/contributions_documentation.html).

mikerob:
Hamish, have installed - thanks.  I'm running on a RasPi so it makes life easier than windows.

Hit something odd, not sure if it's a PEBKAC or a code issue:

from dronekit import connect
from dronekit.lib import APIException

try:
  vehicle=connect('/dev/ttyAMA0', heartbeat_timeout=10)
except APIException as ai:
  print "APIException fail - Something bad happened"
  print type(ai)
  print ai.args
  print ai.msg
except Exception as ai:
  print "Something we didn't expect"
  print type(ai)
  print ai.args
  print ai
else:
  print "It worked"

If I set timeout to 10 or more, it works fine.  If less than that, it throws the following:
pi@MavStation ~/pigs $ python to_exc.py
>>> Link timeout, no heartbeat in last 5 seconds
Exception in thread Thread-1:
  File "/usr/lib/python2.7/threading.py", line 552, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 505, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/dist-packages/dronekit-2.0.0rc7-py2.7.egg/drone
kit/__init__.py", line 77, in mavlink_thread
    fn(self)
  File "/usr/local/lib/python2.7/dist-packages/dronekit-2.0.0rc7-py2.7.egg/drone
kit/lib/__init__.py", line 855, in listener
    raise Exception('>>> No heartbeat in %s seconds, aborting.' % self._heartbea
t_error)
Exception: >>> No heartbeat in 9 seconds, aborting.

APIException fail - Something bad happened
<class 'dronekit.lib.APIException'>
()
Timeout in initializing connection.


It looks like setting timeout too low interferes with or gets inside some other exception handling loop.

tcr3dr:
Hi @mikerob, the default heartbeat_timeout is 10 seconds. You should be able to set it to 0 to indicate you want no timeout. Otherwise it's possible it will throw early on in the connection process if it doesn't receive a heartbeat in time.

The timeout configuration is very new, so possibly buggy! If you have an issue, please file it as a separate issue and we can address it there.

PR #408, opened 2015-11-03
Replaces await_params with wait_ready.

tcr3dr:


tcr3dr:
@hamishwillee Pinging you so you're aware this is merged. This lands after rc5 so it's only on HEAD for now.

Issue #416, opened 2015-11-04
Waiting on arming


tcr3dr:
Considering the GPS and EKF issues exposed by #406, I wonder if we shouldn't be exposing a different parameter indicating "readiness to arm".

In several of our examples, we sidestep this by waiting for a GPS lock. e.g. in simple_goto:

```
    print "Basic pre-arm checks"
    # Don't let the user try to fly autopilot is booting
    if vehicle.mode.name == "INITIALISING":
        print "Waiting for vehicle to initialise"
        time.sleep(1)
    while vehicle.gps_0.fix_type < 2:
        print "Waiting for GPS...:", vehicle.gps_0.fix_type
        time.sleep(1)
```

We discover with 3.3 we also need to wait for an EKF readiness flag. What if we were instead to abstract several of these considerations into an attribute, `vehicle.armable`:

```
vehicle = connect('...', wait_ready='armable')
vehicle.armed = True
```

This would encompass the MODE, GPS, and (depending on vehicle?) EKF checks, so that we can have a simple mechanism to wait for this (instead of our crufty spin-loops).

tcr3dr:
We may also want to expand the `wait_ready` mechanism to encompass matching values.

```
vehicle = connect('...', wait_ready='armable')
vehicle.armed = True
vehicle.wait_condition(armed=True)
```

mrpollo:
what if i have a special condition they would like to meet, can i extend armable?

tcr3dr:
(I actually take back my earlier comments about extending `wait_ready` and `wait_condition`; that would be mixing information synchronization with conditions. Conditional checks should probably remain in user code.)

tcr3dr:
@mrpollo Perhaps, to keep this simple, we can go for something like the following:

```
class Vehicle():
    ...
    @property
    def is_armable(self):
        return GPS and EKF and MODE
    ...
```

And subclasses of Vehicle (using the vehicle_class mechansim) can just override is_armable(). Thus our code for simple_goto becomes:

```
print "Basic pre-arm checks"
while not vehicle.is_armable:
    time.sleep(1)
vehicle.armed = True
while not vehicle.armed:
   time.sleep(1)
```

This is still ugly, but we don't have any general solution yet for "await condition in user code with timeout" and that sounds like a larger, separate issue.

mrpollo:
agreed

hamishwillee:
[...]

PR #423, opened 2015-11-06
Corrects .next definition to be current waypoint (not command entry).

tcr3dr:
Note the 'ekf' bit may go away with #416.


hamishwillee:
@tcr3dr Works great - both the definition of next and also pushing next to the next waypoint.

hamishwillee:
@tcr3dr Added improvements to the example/docs to go with this, and also clarification on ``.next`` to show how it relates to the commands array indexing.

Issue #61, opened 2015-04-09
Attribute observer callbacks pass in the name of the attribute, not the value


eliao:
This seems unexpected to me.  

eg, If I call:
vehicle.add_attribute_observer('mode', mode_callback)

mode_callback is called with 'mode' as a parameter.  I would expect to be passed the value of vehicle.mode.  I'm not sure why I would want the name of the attribute, as I already know the name.  

Perhaps the intention is for a single callback to handle multiple attributes?  That seems odd to me, and the API does not enforce such a construct.

mrpollo:
I agree, this is confusing, right now the right way of doing reading the value is just reading through the global on the callback listener, the value will be updated before the trigger of callbacks relevant source for this behavior is here:

**APIModule**
https://github.com/diydrones/dronekit-python/blob/master/droneapi/module/api.py#L395-L397

**HasObservers**
https://github.com/diydrones/dronekit-python/blob/master/droneapi/lib/__init__.py#L179-L193

I'm not entirely sure why we went this route, but I'll find out and maybe try to add a feature request for next version to also include param values, what would you expect the normal behavior form a callback wold be?

eliao:
I would expect the callback to pass in the new value of the attribute.  

Preferably only when it changes, as referenced by https://github.com/diydrones/dronekit-python/issues/60

tcr3dr:
This is changed as of #426 to have the value passed in.

PR #426, opened 2015-11-06
Attribute handler refactoring

tcr3dr:
Renames some things late in the game:

* renames `on_attribute` to `add_attribute_listener`
* renames `on_message` to `add_message_listener`
* renames `attribute_listener` to `on_attribute`
* renames `message_listener` to `on_message`

This creates more parity between `add_attribute_listener` and `remove_attribute_listener`, which attach and remove functions, and the simple shorthand / decorate `on_attribute` which would be pretty difficult to remove.

* Make `notify_attribute_listeners` and `notify_message_listeners` public

As per request.

* Pass attribute value into attribute listener.

See #61. Function prototype is now: `def listener(vehicle, attr_name, value)`

### DIFF
--- a/docs/examples/create_attribute.rst
+++ b/docs/examples/create_attribute.rst
@@ -166,7 +166,7 @@ the attribute and then notifies all observers.
         self.raw_imu.zmag=message.zmag
         
         # Notify all observers of new message.
-        self._notify_attribute_listeners('raw_imu') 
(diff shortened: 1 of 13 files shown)


hamishwillee:
@tcr3dr Added a bunch of example and docs fixes associated with this change. I was previously a bit confused about the separate add_ vs on_distinction. This is much more comprehensive and also covers use of the wildcard.

PR #429, opened 2015-11-07
Supports cached home_location, replaces default 0,0,0 location with None.

tcr3dr:
* home_location, on setting, now caches value until actual change is set.
* 0,0,0 home location tuple is now ignored. `home_location` will read None until a valid value is read.


hamishwillee:
@tcr3dr Works well/much better. I've added updates to ref, example and guide to cover this. This includes improving the example so that it won't attempt to read or write the home location until a non-None value is detected.

PR #430, opened 2015-11-07
Adds is_armable property.

tcr3dr:
Questions:

* Should we throw if `armed` is set and we are not `is_armamble`?

See #416.

### DIFF
--- a/dronekit/lib/__init__.py
+++ b/dronekit/lib/__init__.py
@@ -1057,6 +1057,12 @@ class Vehicle(HasObservers):
 
     @property
     def is_armable(self):
+        """
+        Returns ``True`` if the vehicle is ready to arm, false otherwise.
+        
+        This attribute wraps a number of pre-arm checks, ensuring that the vehicle has booted,
+        has a good GPS fix, and that the EKF pre-arm is complete.
+        """
         # check that mode is not INITIALSING
         # check that we have a GPS fix
         # check that EKF pre-arm is complete
--- a/dronekit/lib/__init__.py
+++ b/dronekit/lib/__init__.py
@@ -1056,6 +1056,13 @@ class Vehicle(HasObservers):
             self._master.arducopter_disarm()
 
     @property
+    def is_armable(self):
+        # check that mode is not INITIALSING
+        # check that we have a GPS fix
+        # check that EKF pre-arm is complete
+        return self.mode != 'INITIALISING' and self.gps_0.fix_type > 1 and self._ekf_predposhorizabs
+
+    @property
     def system_status(self):
         """
         System status flag according to the MAVLink 
--- a/docs/examples/simple_goto.rst
+++ b/docs/examples/simple_goto.rst
@@ -70,12 +70,19 @@ On the command prompt you should see (something like):
     >>> APM:Copter V3.4-dev (e0810c2e)
     >>> Frame: QUAD
     Basic pre-arm checks
-    Waiting for GPS...: None
-    ...
-    Waiting for GPS...: None
+     Waiting for vehicle to initialise...
+     Waiting for vehicle to initialise...
+     Waiting for vehicle to initialise...
+     Waiting for vehicle to initialise...
+     Waiting for vehicle to initialise...
     Arming motors
      Waiting for arming...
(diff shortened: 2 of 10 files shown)


tcr3dr:
Running on the #430 branch, my script looks like this:

```
python simple_goto.py --connect tcp:127.0.0.1:5760
Connecting to vehicle on: tcp:127.0.0.1:5760
>>> APM:Copter V3.3 (d6053245)
>>> Frame: QUAD
>>> Calibrating barometer
>>> Initialising APM...
>>> barometer calibration complete
>>> GROUND START
Basic pre-arm checks
 Waiting for vehicle to initialise...
 Waiting for vehicle to initialise...
 Waiting for vehicle to initialise...
 Waiting for vehicle to initialise...
 Waiting for vehicle to initialise...
Arming motors
 Waiting for arming...
 Waiting for arming...
 Waiting for arming...
 Waiting for arming...
Adds is_armable check.
>>> ARMING MOTORS
>>> GROUND START
 Waiting for arming...
>>> Link timeout, no heartbeat in last 5 seconds
>>> ...link restored.
>>> Initialising APM...
Taking off!
 Altitude:  0.0
```

Given the GPS + EKF + MODE checks, this seems like a reasonable amount of time. Would you agree @hamishwillee? You can also use the `--speedup 5` parameter in dronekit-sitl to make development faster.

dronekit/lib/__init__.py
```diff
@@ -999,6 +999,13 @@ def armed(self, value):
             self._master.arducopter_disarm()
 
     @property
+    def is_armable(self):
+        # check that mode is not INITIALSING
+        # check that we have a GPS fix
+        # check that EKF pre-arm is complete
+        return self.mode != 'INITIALISING' and self.gps_0.fix_type > 1 and self._ekf_predposhorizabs
```
hamishwillee: I think arming requires a 3D Fix (on Copter at least). Should be checked.

hamishwillee:
> Should we throw if armed is set and we are not is_armamble?

I think so. They really do need to make this check because there is no other way to check on ekf warmup. If we don't force them to check they will have failure for "no obvious reason".

Originally you were talking about publishing a reason so people could work out why things were failing. Decided not to do that? I guess we can publish the checks.

tcr3dr:
I don't want to delve too heavily into "feature development" around this, but to address the reason question, maybe this should be a different function:

```
(armable, reasons) = vehicle.arm_check()
```

Reasons can be a dict of checks that were performed, armable can be a simple boolean value.

In reality, we also do nothing to allow you to read the acknowledgement value which is how we should be doing this—`vehicle.armed = True` should throw if the acknowledgement fails, but we currently don't check for that.

Given that ACKs would be really nice to integrate but is a breaking change and not well thought out, I'm a bit of a loss on what to do here. `arm_check()` and infrastructure for checking arming? Implement ACKs? Implement a second, explicit mechanism `.acknowledge()` that throws if the command failed?

hamishwillee:
@tcr3dr As you say, we could go around this all day. 

I guess what I really want is to be able to have a simple instruction that tells the system to arm that I can wait on until arming is confirmed to succeed. It should have (optional) debug output on what we are waiting on and a time timeout or fail retry timout.

So something like the below, where debug rate is the spacing between printing conditions you're waiting on to console (None for don't print), retry is how many times it will try resend the arm command after EKF warmup is complete. 
```
arm(await=True, retry=3, debug_rate=None)
```
Which returns after arming is detected.

I see the method we have discussed as a primitive for this sort of functionality. Which is why I quite like the ability to get the reasons (though only the ones that are holding us up or we are waiting on).

Does that help? If not, lets KISS - deliver this exactly as currently specced, no throw, no checking.

tcr3dr:
Considering how much work this involves, and that we can do it in a backward-compatible way, I'm going to classify this as "enhancement" and go ahead with is_armable as we've specced. I'd love to develop a more coherent response to awaiting data and acknowledgement (particularly in Python) but that could take some time. No time to lose!

hamishwillee:
@tcr3dr OK. I've added the docs for this, and updated the examples which have a takeoff to use it. Merge at will!

PR #431, opened 2015-11-07
Prevents division error in battery instantiation.

### LINKED ISSUE #298: DPKY2.0 Race condition in Vehicle.battery
The vehicle_state.py example fails with 
```
  File "D:\github\dronekit-python\examples\vehicle_state\vehicle_state.py", line 23, in <module>
    print " Battery: %s" % vehicle.battery
  File "build\bdist.win-amd64\egg\droneapi\module\api.py", line 182, in battery
  File "build\bdist.win-amd64\egg\droneapi\lib\__init__.py", line 181, in __init__
TypeError: unsupported operand type(s) for /: 'NoneType' and 'float'
```
I think this is is due to a race condition. We are getting the Battery object before it is initialised with a real value. The result is a divide by None error because the default value is set to None in the interface. You can reproduce this using code like below ... and if you uncomment the timer before the print for a few seconds the error does not occur.
```
from droneapi import connect
from droneapi.lib import VehicleMode
from pymavlink import mavutil
import time

api = connect('127.0.0.1:14550')
vehicle = api.get_vehicles()[0]
#time.sleep(1)
print " Battery: %s" % vehicle.battery
time.sleep(3)
print 'exit'
```

Note also that on startup all the values are None because they haven't had a chance to initialise with values. I kind of feel we shouldn't return from the vehicle call until we have got some vehicle information ....

tcr3dr:
Fixes #298.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -177,7 +177,7 @@ class MAVHandler:
             time.sleep(0.1)
         self.master.close()
 
-def connect(ip, wait_ready=None, status_printer=errprinter, vehicle_class=Vehicle, rate=4, baud=115200, heartbeat_timeout=30):
+def connect(ip, _initialize=True, wait_ready=None, status_printer=errprinter, vehicle_class=Vehicle, rate=4, baud=115200, heartbeat_timeout=30):
     handler = MAVHandler(mavutil.mavlink_connection(ip, baud=baud))
     vehicle = vehicle_class(handler)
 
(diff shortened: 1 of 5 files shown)


tcr3dr:
This test is now fixed to have an actual test, which proved there was an actual bug in the implementation. Note that the `_initialize` flag to `connect` is private and shouldn't be used / documented yet.

hamishwillee:
@tcr3dr Cool. Unusual to see a private argument to a function.

tcr3dr:
@hamishwillee It is! I can refactor this by splitting out the initialization of the connect() method, but that's lower priority.

Issue #277, opened 2015-08-10
Add improved example for set_mavlink_callback


hamishwillee:
``set_mavlink_callback`` allows users to intercept all messages to the companion computer/script, so it is arguably* quite useful for adding support for messages or "attributes" that aren't yet in the API. The proposal is to update the example to provide a "useful" reference for people who want to create their own message handlers.

* I say "arguably" because it is similar effort to update the API to do this properly. So you would only use this while waiting on changes to go into the API "properly". The example would co-exist with an enhancement to the contribution section providing step-by step guidance on how to support a new attribute. The instructions would make it clear that the best approach is to support the attribute.

hamishwillee:
@tcr3dr @mrpollo @djnugent  I have created an example ([see gist here](https://gist.github.com/hamishwillee/1ff5c3771eea307473c0#file-handle_message-py)) which shows how you might use ``set_mavlink_callback`` to the raw_imu message. The example is a little non-trivial in that it attaches the new object ``vehicle.raw_imu`` to the vehicle and hooks into the normal notifier mechanism so you can create callbacks and, after adding the handler, treat this as though it were an inbuilt attribute.

I would have liked to make this a separate module but there are two problems

* I don't think you can load a module/file that isn't properly in the path - ie you can't import a file "in the same directory" because the current directory is where MAVProxy starts, not where your first file is run from. Yes we could do some magic to add paths but not worth it here unless there ends up being an API to easily set current path from MAVProxy.
* ``set_mavlink_callback`` is a "singleton" - if I hook it in my module, then anyone adding a new observer in their script would break it.

I quite like this, because it is probably a more realistic way you might use the API and because we're not too fast to add new attributes. I dislike it because I would rather encourage people to add new APIs properly.

A question

1. Is this a reasonable example, or should I just do a basic one which simply accesses the values of the received message?

tcr3dr:
I think this mechanism, though less than ideal, will actually go a fair way toward helping us shape what the ideal API looks like for implementing these attributes. So in the case that you add your own parsed messages from MAVLink and then use them in the same example, this is the most common usage of `set_mavlink_callback` even if we don't recommend it—creating globals or modularizing it is simply not possible at this time. When we extend the API though, we can naturally update the same code and hope implementations that used the old methods (which will continue to work) follow.

akwatson:
[...]

PR #437, opened 2015-11-09
Updates Create Attribute example.

tcr3dr:
This addresses some of the questions brought up by #277 and #420.

The goal is "the correct way for users to implement custom attributes". I rewrote this example to leverage the vehicle_class property of connect to instantiate a custom Vehicle class. The Vehicle constructor works similarly to how the internal ones, allowing you to add private properties, message listeners, etc. that become part of the class definition.

I recommend this as the definitive way downstream users implement custom properties, as it's similar to how internal extensions like [dronekit-python-solo](https://github.com/dronekit/dronekit-python-solo) do it.

### DIFF
--- a/examples/create_attribute/create_attribute.py
+++ b/examples/create_attribute/create_attribute.py
@@ -10,10 +10,10 @@ intercepted using the message_listener decorator.
 Full documentation is provided at http://python.dronekit.io/examples/create_attribute.html
 """
 
-from dronekit import connect
-from dronekit.lib import VehicleMode
+from dronekit import connect, Vehicle, VehicleMode
 from pymavlink import mavutil
 import time
+from my_vehicle import MyVehicle
 
 
 #Set up option parsing to get connection string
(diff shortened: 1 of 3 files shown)


tcr3dr:
A key aspect of this approach is that we fully utilize Python's class system, allowing documentation and tooling to react normally. With multiple inheritance, you can support multiple Vehicle "extensions", though I suspect there's no need for more than one custom Vehicle class in a given application.

examples/create_attribute/create_attribute.py
```diff
@@ -10,7 +10,7 @@
 Full documentation is provided at http://python.dronekit.io/examples/create_attribute.html
 """
 
-from dronekit import connect
+from dronekit import connect, Vehicle
```
hamishwillee: Do we really need a model where we separately import from dronekit and dronekit.lib? Would be cool to not have to think about where you import from.

examples/create_attribute/create_attribute.py
```diff
@@ -69,48 +65,56 @@ def __str__(self):
...
+        @self.on_message('RAW_IMU')
+        def listener(self, name, message):
+            """
+            The listener is called for messages that contain the string specified in the decorator,
+            passing the vehicle, message name, and the message.
+            
+            The listener writes the message to the (newly attached) ``vehicle.raw_imu`` object 
+            and notifies observers.
+            """
+            self.raw_imu.time_boot_us=message.time_usec
```
hamishwillee: These should all be ``self._raw_imu`` right? (i.e. with the private member)

examples/create_attribute/create_attribute.py
tcr3dr: Oh, this should change only to `from dronekit import ____`. Didn't notice the `.lib` part.

examples/create_attribute/create_attribute.py
tcr3dr: Yes.

hamishwillee:
OK, looks really great. I think people are going to want to potentially create their own libraries for re-usability. Can you split into two files if you agree? 

Bonus points if we can accept files of the right format as a "plugin" to our libraries - ie so that we can take an external file of the right format and dump it into our source tree and have everything work automatically :-) 

Drop me a line when you're done (whatever you decide should be done) and I'll look at a docs update.

tcr3dr:
@hamishwillee 

> Can you split into two files if you agree?

To clarify, separate the Vehicle class definition from the example? I can do that.

hamishwillee:
>> Can you split into two files if you agree?

> To clarify, separate the Vehicle class definition from the example? I can do that.

Yes please, and fix up the other minor bugs raised above.

tcr3dr:
@hamishwillee Bumped.

hamishwillee:
@tcr3dr Thanks for making the fixes. I think this is a really good example now - both clear and useful. I've added docs for it, and also removed a couple of unnecessary imports in both files. Should be ready to merge.

PR #440, opened 2015-11-09
Fixes drone delivery example.

tcr3dr:
The original example was starting the web server on the callback thread, ensuring no messages could be processed again. This was what was breaking.

### DIFF
--- a/examples/drone_delivery/drone_delivery.py
+++ b/examples/drone_delivery/drone_delivery.py
@@ -40,52 +40,14 @@ cherrypy_conf = {
 }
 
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Looks good to me, though you may want to update to our new is_armable takeoff code once both are in.

tcr3dr:
@hamishwillee Maybe review this one again when you get the chance before I merge it in. I made some changes, removed some dead code.

hamishwillee:
@tcr3dr I didn't get round to rechecking this, but I hope to ... next.

PR #441, opened 2015-11-10
Implements SystemStatus property.

tcr3dr:
Replaces #156.

This adds a `SystemStatus` property which functions similarly to `VehicleMode`, in that it is a light wrapper around a string (really an enum).

This is a breaking change with earlier rc's in that it used to be a number from mavutil, and is no longer.


hamishwillee:
@tcr3dr Looks good. I made a very minor change so that the example code reads ``vehicle.system_state.state`` - slightly prettier than printing ``vehicle.system_state`

PR #442, opened 2015-11-10
Allows parameters to be iterated and observed.

tcr3dr:


hamishwillee:
Looks good, but I don't want to play with the docs for this because I've made quite a lot of changes to another PR. So I will update docs for this after both this and that PR have been merged.

hamishwillee:
@tcr3dr I haven't updated the test or documentation, but basic testing on this functionality indicates that it works as indicated.

A few "semi interesting" things discovered. 

1. The observer is case-sensitive - it will detect change to THR_MIN if you set THR_MIN. Since I thing (?) that ardupilot doesn't care that is a bit of a gotcha
1. The time taken for the callback to be called seems to be between 1 and 4 seconds after setting.
1. The result isn't cached. NOt a big deal, but wouldn't it make sense to do so?

tcr3dr:
1. Changed this to not be case-sensitive.
2. Not sure how to address speed issues, at least in the short term.
3. We now cache parameter results. In practice I expect this to be unnecessary—but semantically if the parameter is unchanged it should not fire again, so better consistent than sorry.

hamishwillee:
@tcr3dr Those changes sound great. Can you merge this, and I'll update the docs from master?

tcr3dr:
@hamishwillee Done.

hamishwillee:
These changes documented in #450

PR #443, opened 2015-11-10
Hgw params etc

hamishwillee:
This fixes a bunch of outstanding minor items.


hamishwillee:
@tcr3dr ONe important thing to note is that in __init__.py the docs MUST be above the module imports. Otherwise Sphinx does not include those comments. I made that change before but it got lost. This update makes the change again. Please don't undo it!

Issue #432, opened 2015-11-07
Setting connection timeout too short (<10s) throws unhandled exception


mikerob:
I was testing out the connection timeout, and came across an odd behaviour.  If I set the timeout > 10 seconds, everything works nicely - it hapily throws an APIException that I can catch.

However, if I set the connection timeout lower than that, it throws a different exception, seemingly from somewhere "inside" the connection, and it's an "Exception" rather than an "APIException".

Is this the intent?  I reported it here: https://github.com/dronekit/dronekit-python/pull/407 and was requested to submit it as a new issue.

from dronekit import connect
 from dronekit.lib import APIException

try:
 vehicle=connect('/dev/ttyAMA0', heartbeat_timeout=10)
 except APIException as ai:
 print "APIException fail - Something bad happened"
 print type(ai)
 print ai.args
 print ai.msg
 except Exception as ai:
 print "Something we didn't expect"
 print type(ai)
 print ai.args
 print ai
 else:
 print "It worked"

If I set timeout to 10 or more, it works fine. If less than that, it throws the following:
 pi@MavStation ~/pigs $ python to_exc.py

 

Link timeout, no heartbeat in last 5 seconds
 Exception in thread Thread-1:
 Traceback (most recent call last):
 File "/usr/lib/python2.7/threading.py", line 552, in bootstrap_inner
 self.run()
 File "/usr/lib/python2.7/threading.py", line 505, in run
 self.__target(self.__args, *self.__kwargs)
 File "/usr/local/lib/python2.7/dist-packages/dronekit-2.0.0rc7-py2.7.egg/drone
 kit/__init.py", line 77, in mavlink_thread
 fn(self)
 File "/usr/local/lib/python2.7/dist-packages/dronekit-2.0.0rc7-py2.7.egg/drone
 kit/lib/init.py", line 855, in listener
 raise Exception('>>> No heartbeat in %s seconds, aborting.' % self._heartbea
 t_error)
 Exception: >>> No heartbeat in 9 seconds, aborting.
 
APIException fail - Something bad happened
 
()
 Timeout in initializing connection.

It looks like setting timeout too low interferes with or gets inside some other exception handling loop.

tcr3dr:
Thanks for filing this, @mikerob. This is definitely a weird case for exceptions—one being throw on the MAVLink thread, another interrupting the main thread—so let me put together a proposed solution for this that is catchable.

hamishwillee:
Hi @mikerob 

[...]

Issue #414, opened 2015-11-04
No timeout on "connect"


mikerob:
It appears that there's no timeout for the "connect" function.  If you try to connect to a serial port that exists but is not actually connected to a vehicle, it just hangs.

On the RasPi, I'm conecting to the Pixhawk via USB, so it's /dev/ttyACM0. The RasPi has a built in serial port at ttyAMA0. If you "Connect('/dev/ttyAM0')" it hangs, presumably because it's a valid port so it thinks it can read, but never gets valid data.

mikerob:
In dronekit-python / dronekit / lib / __init__.py it calls "wait_heartbeat", which doesn't timeout (mavutil.py - lines 326 and 372)

I'm not sure what the solution is, other than to write a new "wait_heartbeat_timeout" function that calls recv_match from mavutil and passes it a timeout value, rather than using the ready-made wait_heartbeat.

# Wait for first heartbeat.
        while True:
            try:
                self._master.wait_heartbeat()
                break
            except mavutil.mavlink.MAVError:
                continue

tcr3dr:
@mikerob,

Thanks for pointing this out. I think this can tie pretty well into creating actual connection timeouts (rather than intermittent dropoffs), which there is a branch exploring now: https://github.com/dronekit/dronekit-python/pull/407

The same connection logic which says after 30s, close the connection and throw, can be used on initialization for the same purpose.

That branch is a WIP but I want to land it soon, as well as making these timeouts (and their verbosity) configurable.

mikerob:
Sounds like there's work already underway.  I'll take the liberty of running through what I'm doing with this, because I have a number of issues - some of them may be relevant to a wider audience.

Overall, I'm working on a lightweight ground station based on a RasPi and 20x4 character LCD (a bit like an updated ArduStation).  I'd like it to be flexible and hands-free (and headless) in terms of connection, so ideally could "automatically" detect how a vehicle is connected.  To that end, I'd like to be able to try network, USB radio, USB Pixhawk, and direct GPIO serial connections.  

1) Managed / more graceful handling of connections that don't exist - would be nice to have the connection return a softer error saying the IP or serial port does not exist. This would make it easier to try different connections.

2) The primary issue rasied above - valid serial port, but no incoming data

3) Managing loss of connection. As I see it there are two different loss of connetion cases:
[...]

PR #447, opened 2015-11-11
Adds smarter error notification when errors occur.

tcr3dr:
This changes APIException to have `.message` not `.msg`, aligning it with all Python error classes.

This also addresses #432 by being less noisy about errors on MAVLink parsing thread.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -6,6 +6,7 @@ import os
 import platform
 import re
 import dronekit.lib
+from dronekit.lib import APIException
 from dronekit.util import errprinter
 from pymavlink import mavutil, mavwp
 from Queue import Queue, Empty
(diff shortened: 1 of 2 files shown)


hamishwillee:
@tcr3dr Don't you also need to update the APIException definition in dronekit.lib for to also have .message?

I don't think this has a documentation implication unless you think we should maintain a list of common exceptions and how to debug them?

tcr3dr:
@hamishwillee Probably no documentation impact. This code now works, which we may want to document at some point (but right now requires cross platform testing, i.e. the os.exception bit only was tested on OS X): https://github.com/dronekit/dronekit-python/issues/414#issuecomment-155658984

hamishwillee:
@tcr3dr OK. The only other reason we might document is to suggest wrapping connect calls in a try/catch block in our "programming guide".

Issue #400, opened 2015-11-03
send_mavlink() doesn't autofix system id


juanmacrack93:
When sending mavlink messages created with message_factory, the correct system id must be specified, otherwise it won't work.

This was tested using the following code:

```python
 msg = vehicle.message_factory.mount_control_encode(
        0, 0,  # target system, target component
        30 * 100,  # pitch
        0,  # roll
        0,  # yaw
        0)  # save trim
vehicle.send_mavlink(msg)
```

However, changing the system id to the proper value (1 in this case) fixes the problem.
The documentation suggests that this should be automatically taken care of by send_mavlink(), yet it does't seem to be the case.

I tried the same code with dronekit 1 and it does indeed work without the need to specify the sysid manually.

tcr3dr:
Hi @juanmacrack93,

Thanks for filing the issue. Message correction is added back in with the refactor on this PR: https://github.com/dronekit/dronekit-python/pull/401

Exposing a way to change the target system and component is not on that PR yet, but should be added in shortly.

tcr3dr:
You can follow #448 for a way to modify the system/component directly.

tcr3dr:
Further addressed by #461.

tcr3dr:
Merged.

PR #448, opened 2015-11-11
Exposes target_system on connect(). Exposes MAVConnection.

tcr3dr:
See #400.

```py
connect('tcp:...:...', target_system=0, target_component=100)
```

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -40,7 +40,7 @@ class MavWriter():
         os._exit(43)
 
 class MAVHandler:
-    def __init__(self, master, vehicle_class=Vehicle):
+    def __init__(self, master, vehicle_class=Vehicle, target_system=0):
         self.vehicle_class = vehicle_class
 
         self.master = master
@@ -51,8 +51,7 @@ class MAVHandler:
         self.master.mav = mavutil.mavlink.MAVLink(MavWriter(self.out_queue), srcSystem=self.master.source_system, use_native=False)
 
         # Targets
-        self.target_system = 0
-        self.target_component = 0
+        self.target_system = target_system
 
         # Listeners.
         self.loop_listeners = []
@@ -151,8 +150,6 @@ class MAVHandler:
         """Set correct target IDs for our vehicle"""
         if hasattr(message, 'target_system'):
             message.target_system = self.target_system
-        if hasattr(message, 'target_component'):
-            message.target_component = self.target_component
 
     def forward_loop(self, fn):
         """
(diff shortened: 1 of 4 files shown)


hamishwillee:
@tcr3dr Can you confirm that parameters are the system/component ID that we are connecting to, and that the plan is to use these when sending messages as in #400 rather than whatever value the user puts in:
```
 msg = vehicle.message_factory.mount_control_encode(
        0, 0,  # target system, target component
```
Is that correct? Are the 0 values above just ignored?

My feeling is that this isn't quite right. 

I think if you have a function like the above which you are explicitly calling it should obey the system/target you enter. If we use the function internally in an exposed API (e.g. goto) then we *would* pass in our target_system value because this is the target that the vehicle represents.

Things are different for target_component because a vehicle represents a particular system, not a particular system and component. Do we really want to switch vehicle to talk to a camera mount (say)?  I suspect that we will need to use broadcast ID to talk to components and hope that the message gets forwarded appropriately. When sending messages "manually" as above, I would assume the coder would think about which component ID they want to direct it to, and we shouldn't override that (just advise use of the broadcast id)

Last of all, we also  want DroneKit to be able to change its **own system_id** so that it is possible for multiple dronekit systems to co-exist and communicate with each other. I am not sure if DK also needs to be able to set its own component ID, but let's pretend a use case exists ....

Hope some of that makes sense. As I said, I may have got this completely wrong.

tcr3dr:
@hamishwillee,

This PR now only allows the specification of the target_system ID. This is all that gets overwritten in MAVLink commands sent using the vehicle object.

If you deliver messages directly to the vehicle `send_mavlink` command, the `target_system` still gets overwritten. I'm unsure what to do here; I feel like we should expect the `target_system` to be fixed to that of the vehicle it's being called on. If we want to communicate with other target_systems, it should be done through (perhaps) another interface?

We should support the changing of our own system's system_id, but I am not sure how that logic works yet, and seems like it can be an enhancement.

hamishwillee:
@tcr3dr Isn't the target_system id the ID of the system you want to talk to? In that case, wouldn't you be better off querying that from the vehicle you've connected to rather than trying to set it in connect?. I presume this is what we do now but maybe we just always broadcast.

Once we have the value, however we get it, then yes, we should use it for our internal message sending since ``vehicle`` represents a specific vehicle.

My (possibly idiotic) take on communicating with other systems is that if you want to do this you should create a separate vehicle object and handle any inter-vehicle communication in your script. The implication of this is

* Yes, the ``target_system_id`` should be used for all communication from this vehicle object (overwriting whatever might be set in ``send_mavlink``.
* IMO if we aren't allowing people to set this parameter in ``send_mavlink``, we should probably remove it from the public API (and if we call the same API internally, we'll need both an internal and external message sending API)

Yes, changing our own system_id is an enhancement. I don't *think* this is complicated - and in most cases it is something you wouldn't need to do. I'll leave you to create a separate issue because I'm a little lazy!

hamishwillee:
@tcr3dr 

I still think it doesn't make much sense to allow people to specify the target_system - this should be obtained from the connected vehicle.  

That said, can you confirm that target_system parameter is only thing requiring docs from this?

tcr3dr:
@hamishwillee See #458. It makes sense in that we actually have a manual override in the Solo code—also this is the first step to implementing it as an automatic process.

`target_system` is the only thing requiring docs. I may or may not be exposing MAVConnection but it probably involves another PR and me writing up some examples about that. For now, ignore.

PR #450, opened 2015-11-12
Hgw parameter docs

hamishwillee:
This updates docs and examples for parameter observing and iteration ( #442). It also documents ``wait_ready`` behaviour in the vehicle class and connect method. 

The changes are recorded in the migration guide which is now linked from the home/index page. I also added some missing doc here for the changes to missions. 

I think it is all sane.Getting weary though!

@tcr3dr please review and merge!

### DIFF
--- a/docs/examples/vehicle_state.rst
+++ b/docs/examples/vehicle_state.rst
@@ -49,32 +49,36 @@ On the command prompt you should see (something like):
 
 .. code:: bash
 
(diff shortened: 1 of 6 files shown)


tcr3dr:
@hamishwillee Looks good. Make one small modification, but we are set.

hamishwillee:
@tcr3dr Thanks very much . Might be worth rebuilding the docs on our site?

PR #457, opened 2015-11-16
Adds last_heartbeat property.

tcr3dr:


hamishwillee:
@tcr3dr  I'm far more interested a property for the "last heartbeat timer value" than the last heartbeat timestamp received. With what we have now I think most users will have to implement their own timer ... which is a bit of a pain.

What I think people will want to do is monitor the timeout timer value and in the callback do something when it reaches a particular value. 

The difficult bit is deciding what is a reasonable update period on the timer. I think we want it to update when the value is set to zero. We also want it to update at least once every second. The question is, whether a more frequent period (tenth of second? would ever be useful).

@mikerob - thoughts, as a real user?

hamishwillee:
@tcr3dr Can't test - immediately get exception:

```
  File "E:\deleteme\dronekit-python\examples\simple_goto\test2.py", line 14, in <module>
    vehicle = connect(args.connect, wait_ready=True)
  File "build\bdist.win-amd64\egg\dronekit\__init__.py", line 190, in connect
  File "build\bdist.win-amd64\egg\dronekit\lib\__init__.py", line 1460, in initialize
dronekit.lib.APIException: Timeout in initializing connection.
```

tcr3dr:
@hamishwillee What source code are you running. I can't recreate, and all tests are passing.

hamishwillee:
This file

```
from dronekit import connect
import time


#Set up option parsing to get connection string
import argparse  
parser = argparse.ArgumentParser(description='Print out vehicle state information. Connects to SITL on local PC by default.')
parser.add_argument('--connect', default='127.0.0.1:14550',
                   help="vehicle connection target. Default '127.0.0.1:14550'")
args = parser.parse_args()

# Connect to the Vehicle
print 'Connecting to vehicle on: %s' % args.connect
vehicle = connect(args.connect, wait_ready=True) 

@vehicle.on_attribute('last_heartbeat')   
def listener(self, attr_name, value):
    # `attr_name` is the observed attribute (used if callback is used for multiple attributes)
    # `attr_name` - the observed attribute (used if callback is used for multiple attributes)
    # `value` is the updated attribute value.
    print "HEART ATTR: %s changed to: %s" % (attr_name, value)
 
   
print ' Sleeping for 10'
time.sleep(10)

#Close vehicle object before exiting script
print "Close vehicle object"
vehicle.close()
```

hamishwillee:
@tcr3dr Documentation added for last_heartbeat. Can now be merged.

PR #460, opened 2015-11-17
updates target_system documentation

hamishwillee:
This provides docs for #448. It updates  target_system information in ``connect()``, and also the send_mavlink information that relates to it. It tidies some of the text and improves the linking from reference to guide.

@tcr3dr This states that the vehicle target system is used to set the values used in ``Command`` and in ``send_mavlink``. 
It states that the target_component should be set to zero by default - the broadcast ID. This is different than previoulsy, in which we said the component id was updated by dronekit.
Please confirm that is OK.

### DIFF
--- a/docs/guide/copter/guided_mode.rst
+++ b/docs/guide/copter/guided_mode.rst
@@ -133,7 +133,10 @@ The command can be interrupted by a later movement command. When moving the vehi
 
 .. tip::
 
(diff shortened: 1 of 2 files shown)


tcr3dr:
Uh, we might have to redo or merge any changes into #461. Especially because we don't expose target_system directly on connect() anymore :)

hamishwillee:
Indeed. Grrr. Can you merge this first, and then I'll fix up docs after #461 goes in?

PR #461, opened 2015-11-18
Implements target_system lookup. Implements better target semantics.

tcr3dr:
Exposing a `target_system` on `connect()` is probably not what we want. Exposing `source_system` might be, though.

This does a bit of an overhaul over the MAVLink object in the backend to make sure we get this right. The `target_system` is loaded from the first heartbeat we get and fixed after that point. We also have some backend classes which we might expose eventually, but for now demonstrate the use of multiple target systems in one program. e.g.

```
dronekit-sitl copter
```

In one window,

```py
import time
from dronekit import connect
from dronekit.mavlink import MAVConnection

v = connect('tcp:127.0.0.1:5760', wait_ready=True, source_system=255)

o = MAVConnection('udpout:127.0.0.1:14550', source_system=254)
v._handler.pipe(o)
o.start()

while True:
	time.sleep(1)
```

in the next, and 

```
mavproxy.py --master 0.0.0.0:14550 --source-system=254
```

Works to daisy chain a connection _through_ DroneKit. This is obviously using undocumented features (MAVConnection directly) but demonstrates the logic should work if we expose it publicly later.

### DIFF
--- a/docs/guide/copter/guided_mode.rst
+++ b/docs/guide/copter/guided_mode.rst
@@ -199,11 +199,12 @@ message fields as arguments:
     # send command to vehicle
     vehicle.send_mavlink(msg)
 
-There is no need to specify the ``target_system`` id in messages (just set to zero) as DroneKit will automatically 
-update messages with the vehicle's internal ``target_system`` (see :py:func:`connect() <dronekit.lib.connect>`) before sending. 
-The ``target_component`` should be set to 0 (broadcast) unless the message is to specific component. 
-CRC fields and sequence numbers (if defined in the message type) are automatically set by DroneKit and can also 
-be ignored/set to zero.
+If a message includes ``target_system`` id you can set it to zero (DroneKit will automatically 
+update the value with the correct ID for the connected vehicle). Similarly CRC fields and sequence numbers 
+(if defined in the message type) can be set to zero as they are automatically updated by DroneKit.
+The ``target_component`` is not updated by DroneKit, but should be set to 0 (broadcast) unless the message is 
+really intended for a specific component. 
+
 
 .. _guided_mode_how_to_send_commands_command_long:
 
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -25,8 +25,8 @@ LocationGlobal = dronekit.lib.LocationGlobal
 LocationLocal = dronekit.lib.LocationLocal
 CloudClient = dronekit.lib.CloudClient
 
-def connect(ip, _initialize=True, wait_ready=None, status_printer=errprinter, vehicle_class=Vehicle, rate=4, baud=115200, heartbeat_timeout=30, target_system=0):
-    handler = MAVConnection(ip, baud=baud, target_system=target_system)
+def connect(ip, _initialize=True, wait_ready=None, status_printer=errprinter, vehicle_class=Vehicle, rate=4, baud=115200, heartbeat_timeout=30, source_system=255):
+    handler = MAVConnection(ip, baud=baud, source_system=source_system)
     vehicle = vehicle_class(handler)
 
     if status_printer:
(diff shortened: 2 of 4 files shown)


tcr3dr:
Another proposal is that `vehicle.send_mavlink` fixes the target_system, but `vehicle._handler.send_mavlink` would be a "raw" write on the connection. We could also expose this as `vehicle.connection` instead of `_handler`, but I don't want to commit too early (as in, we can do this after 2.0.0)

hamishwillee:
My first reaction (second to follow)

- ``target_system`` is the MAVLink network ID of the vehicle our script is talking to. AFAIK most vehicles set themselves as 1 by default. We can set this to 0 to talk to everything, but we should set it to whatever the vehicle thinks its id is. 
- ``source_system`` is the id of the script on the MAVLink network. This ID is what vehicles will use to send messages back to us, and needs to be unique. As a result it needs to be settable.

UPshot, 

- yes, removing target_system from the connect API and getting the id from the connected vehicle makes sense (I can't think of a use case where we would want to set this to anything other than what the connected vehicle says it is). NOTE - the vehicle can change its ID, so we may need to monitor for this?

- Yes we need  ``source_system`` to be set for the same reason - to enable our DK instructions to be uniquely identified. This actually does matter because there are settings on the autopilot where the autopilot will only take guided instructions from a specific GCS/id.

hamishwillee:
@tcr3dr Second thoughts ...

1. Why is this approach for setting up connections to multiple  targets within a script better than creating a second vehicle class and connecting it to that target? 

1. Some of our customers will be very happy (based on what they've asked) because I think what you've done here is enable automated forwarding of all packets to a specified address. ie MAVProxy in dronekit. Is that correct?

1. Not sure how this would look in a real app - will need more examples if you want this documented.

As an aside, the two reasons I  wanted UDP was to be able to use the loopback address (simplify instructions) and to have consistency with main instructions on old SITL. I did test tcp connection to sitl via loopback at one point and it failed. However it seems to work now, so I'm thinking perhaps just using tcp is fine for the examples.  

Are there other reasons why we might want UDP?

W.r.t the example, did you mean?
```
mavproxy.py --master 0.0.0.0:14550 --target-system=254
```

tcr3dr:
@hamishwillee Using SITL with TCP means for more connection overhead we gain predictability. I'm inclined to want this in most examples; for real-world testing, UDP is more practical, but so are a number of other simulation variables (wind speed, battery depletion, real physics, etc.) Since examples are meant to be short and understandable, I prefer TCP.

> NOTE - the vehicle can change its ID, so we may need to monitor for this?

We would! I'd want to be able to test this before adding it in, though, so I can say post-2.0.0 we should get back to it. 

> Why is this approach for setting up connections to multiple targets within a script better than creating a second vehicle class and connecting it to that target?

If you want two vehicle objects, you should connect to the same master. This particular scenario is for if you want to create i.e. udpout connections. Right now we have the scaffolding in place that will allow that down the line. 

> Some of our customers will be very happy (based on what they've asked) because I think what you've done here is enable automated forwarding of all packets to a specified address. ie MAVProxy in dronekit. Is that correct?

Yup. Right now I have a DK script that forwards to udpout:127.0.0.1:14550 for the Solo app in emulation. Works well!

> Not sure how this would look in a real app - will need more examples if you want this documented.

Let's just document the source_system param, not the rest of this. You can still connect multiple vehicles to the same source address, which was necessary.

hamishwillee:
Hi @tcr3dr 

That all makes sense - and agree. I will assume you'll create a tracking issue on watching for system id changes.

Can you merge #461 first and then merge this on top of it? I can update the docs fairly easily after that (and more easily) than starting fresh. If that is too hard, I'll take a shot at merging fresh - just confirm which is best.

Cheers
H

tcr3dr:
Done.

hamishwillee:
@tcr3dr Docs updated now. Please check the statements just on the connect() method for heartbeat_timeout and source_system. Mostly I want to check that ``heartbeat_timeout`` only applies to the connection - or is this the same timeout value used to monitor the link for activity after connection?

tcr3dr:
This looks good! `heartbeat_timeout` applies to the connection even after heartbeat is established. (it may be the case that we don’t want this behavior, but right now that’s what it does)

PR #465, opened 2015-11-18
dronekit.lib -> dronekit and final code reformatting.

tcr3dr:
* dronekit.lib -> dronekit (dronekit.lib exists still as an alias)
* dronekit.lib.CloudClient -> dronekit.cloud.CloudClient
* formatting using yapf definition (minor whitespace mostly)


tcr3dr:
@hamishwillee Gonna go ahead with this one, assuming we'll need one last PR for doc finishing etc.

PR #469, opened 2015-11-19
Fixes vehicle.commands.goto with LocationGlobal.

### LINKED ISSUE #467: Vehicle.commands.goto always treats alt as relative
Vehicle.commands.goto takes a LocationGlobal or a LocationGlobalRelative but either way it assumes the alt value is relative. This means that if you pass it a LocationGlobal and say "goto" it will fly to that position ABOVE the home position (rather than mean sea level). 

Internally we need to detect we're dealing with a global and pass in the associated relative alt.

tcr3dr:
Fixes #467.

In the way `.goto` is currently implemented, APM just ignores information about relative vs absolute location and defaults to absolute. This can be considered a bug, [or our use of MISSION_ITEM can be considered a hack](https://groups.google.com/forum/#!topic/drones-discuss/_UmqPM5b_60), and either way, we want a solution that works yesterday. For now, this restores the expected behavior, with minimal side effects (hopefully).

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -2110,15 +2110,24 @@ class CommandSequence(object):
         '''
         if isinstance(location, LocationGlobalRelative):
             frame = mavutil.mavlink.MAV_FRAME_GLOBAL_RELATIVE_ALT
+            alt = location.alt
         elif isinstance(location, LocationGlobal):
-            frame = mavutil.mavlink.MAV_FRAME_GLOBAL
+            # This should be the proper code:
+            # frame = mavutil.mavlink.MAV_FRAME_GLOBAL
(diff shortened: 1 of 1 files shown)


hamishwillee:
@tcr3dr Looks good to me by inspection.

I suspect this may be a case where Copter and Plane behave differently. In any case, your solution standardises the behaviour. 

This might be a case where another MAVLink autopilot could respect the frame. I wonder if there is a) any benefit to respecting the frame b) if so, any way to add markers of "things to check for on other autopilots.

Either way, happy for you to merge.

hamishwillee:
PS The last push I did either fixed up everything to a LocationGlobalRelative or it didn't matter. There are no docs issues with this change.

PR #471, opened 2015-11-19
Update vehicle_state.py

franknitty69:
vehicle.home_location requires a float, not an integer. changed my_location_alt.alt value from 222 to 222.0

### DIFF
--- a/examples/vehicle_state/vehicle_state.py
+++ b/examples/vehicle_state/vehicle_state.py
@@ -65,7 +65,7 @@ print "\nSet new home location"
 # Home location must be within 50km of EKF home location (or setting will fail silently)
 # In this case, just set value to current location with an easily recognisable altitude (222)
 my_location_alt=vehicle.location.global_frame
-my_location_alt.alt=222
+my_location_alt.alt=222.0
 vehicle.home_location=my_location_alt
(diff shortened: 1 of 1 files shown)


tcr3dr:
@franknitty69 Thanks for the PR! I'll make sure this passes CI and pull it in.

Would it be useful if the Location{Global,GlobalRelative,Local} objects coerced their attributes into floats when they were assigned? Thus `.alt = 5` would read back as `.alt == 5.0`. This may lead to less user error, though it seems a bit magical.

hamishwillee:
@franknitty69 What was the impact of not doing this? The code ran fine without crashing for me and seemed to work.

franknitty69:
i setup a new vagrant instance and ran the vehicle_state example. it failed with an error on line 69. the error complained about the struct parameter not being a float. when i changed the alt from 222 to 222.0 the error went away. the funny thing is that now if i change it back, i can no longer recreate the error.

hamishwillee:
That is odd. Either way, it should be robust enough to cope. I'm sure @tcr3dr will find some way to make it bullet proof :-)

tcr3dr:
Pulled in by #477!

PR #474, opened 2015-11-20
Hgw test fix remaining examples

hamishwillee:
This fixes up all docs and example issues found after re-running the examples against master.


hamishwillee:
@tcr3dr this can be merged after your review. FYI I think the easiest issues to miss are those when listening for a location object, since what is returned is a Locations object rather than a specific location type.

tcr3dr:
Reviewed, lgtm

PR #476, opened 2015-11-23
Hgw more fixes for goto takeoff

hamishwillee:
@tcr3dr This is last round of fixes to docs/examples to address the changed goto/takeoff API. Fixes a fairly significant bug in the guided mode example. 

I think ready to release now after release notes done and link added to v1 docs.


tcr3dr:
This looks good. My only comments are in the custom functions (instances vs type equality checking), which I suppose we will eventually roll into the library. So in the interest of time...

hamishwillee:
Great. Best to make the library as robust as possible.

Because we publish "master" as the right place to get the examples they really have to be "always runnable". Hence the fix done.

PR #479, opened 2015-11-23
Defaults vehicle_class to None, not Vehicle.

tcr3dr:
Future-proofing.


hamishwillee:
@tcr3dr I assume we still don't have to pass Vehicle in explicitly when we connect?

tcr3dr:
@hamishwillee Yup.

Made this change so in case we start emitting CopterVehicle, AirplaneVehicle, RoverVehicle by default, we aren't tied to the generic Vehicle class from connect().

hamishwillee:
Great.

PR #483, opened 2015-11-26
Improve information about setting baud rate when connecting

hamishwillee:
Apparently connecting to autopilot via 3DR Telementry radio requires a different baud rate too. I've added this use case to the table, and a "general" note about the possible need to set baud rates.

One issue is that the user asserted a string that included the baud rate (rather than using the baud parameter we supply to connect() ). I'm asking for confirmation about what actually works on 2.0.0. Please don't merge until that is confirmed.

### DIFF
--- a/docs/guide/getting_started.rst
+++ b/docs/guide/getting_started.rst
@@ -137,7 +137,7 @@ the more common connection types:
    * - Linux computer connected to the vehicle via USB
      - ``/dev/ttyUSB0``
    * - Linux computer connected to the vehicle via Serial port (RaspberryPi example)
-     - ``/dev/ttyAMA0, baud=57600``
+     - ``/dev/ttyAMA0`` (also set ``baud=57600``)
    * - SITL connected to the vehicle via UDP
      - ``127.0.0.1:14550``
    * - OSX computer connected to the vehicle via USB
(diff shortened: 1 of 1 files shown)


hamishwillee:
@tcr3dr The person who reported the issue did some testing against 2.0.0

The modified vehicle_state.py and found:

* vehicle = connect('com7',baud=57600, wait_ready=True)  # Recommended way - WORKED
* vehicle = connect('com7,baud=57600', wait_ready=True)  # DID NOT WORK

I have updated the docs to separate out the baud parameters in the connection string area.

What confuses me is that when they *passed in* a string that is equivalent to the "did not work" case above then it did work:

```
python vehicle_state.py –connect com7,57600
```

Any ideas why? My code doesn't split out the baud rate and pass it in.

Anyway, this is good to merge ... unless the above highlights a serious issue and/or you think I should add baud rate to all the examples too?

tcr3dr:
@hamishwillee In part this may depend on the Windows shell used. For example, Powershell interprets commas as separate arguments (which is insane). If they're using CMD.exe, it's different. Can you find out?

hamishwillee:
@tcr3dr I asked, but did not get an answer I can use. In either case I don't think it matters - this change can be merged because we don't "support" the connection string including baud rate approach, even if it does happen to work, or happen to work on some shells/platforms.

PR #491, opened 2015-12-01
drone_delivery fails when try to execute go to command, this is due t…

GOPINATH-GS4:
…o a bug executing simple_goto

changed line 113 and 120 from self.simple_goto to self.vehicle.simple_goto


tcr3dr:
Pulled in via #492. Thanks a ton @GOPINATH-GS4!

PR #497, opened 2015-12-10
Gimbal control

djnugent:
Adds gimbal control
* rotate(pitch roll yaw) 
* target_gps(roi)
* release(): give back rc control of gimbal
* attributes: current pitch roll yaw

Documentation may need some work. And we may want to add type checks

I apologize for the messed up commit. It appears my text editor went to town on white spaces.


djnugent:
I also need verification on the attribute notifier stuff.

dronekit/__init__.py
```diff
@@ -1893,6 +1884,115 @@ def wait_ready(self, *types, **kwargs):
 
         return True
 
+class Gimbal(HasObservers):
+    """
+    Gimbal control and status.
+
+    An object of this type is returned by :py:attr:`Vehicle.gimbal`.
+    .. note::
```
hamishwillee: You may well need a line space here for the note to render.

dronekit/__init__.py
```diff
@@ -1893,6 +1884,115 @@ def wait_ready(self, *types, **kwargs):
...
+        self._vehicle.send_mavlink(msg)
+        msg = self._vehicle.message_factory.mount_control_encode(
+                    0, 1,    # target system, target component
+                    pitch * 100, # pitch is in centidegrees
+                    roll * 100, # roll
+                    yaw * 100, # yaw is in centidegrees
+                    0) # save position
+        self._vehicle.send_mavlink(msg)
+
+    def target_gps(self,roi):
```
hamishwillee: This probably should support either type of global frame. For brownie points, would be great if it could also take a LocationLocal. @tcr3dr Pity we don't support the body frame for local, because it would be very handy to be able to say "set ROI 10metres in front of me"

hamishwillee:
@djnugent Looks good to me (by inspection). I can probably test this tomorrow. Any way you can fix that commit - hard to work out what is new :-( I can fix up the docs.

Aside from my point above, are there any requirements on this code? For example, I am presuming that this will work on any gimbal/vehicle with gimbal and that if you have no gimbal the commands will set the ROI but otherwise do nothing?

tcr3dr:
Fixed whitespace issues, tests, filed as #498.

Didn't address @hamishwillee's comments yet.

hamishwillee:
Great. #498 easier to read. I guess maybe we need another position type - relative to the local frame to address the case I refer to above.

tcr3dr:
I'm going to close this in favor of #498, since it's passing tests.

PR #498, opened 2015-12-10
Gimbal: Added gimbal control

tcr3dr:
Merges #497.

### DIFF
--- a/docs/examples/vehicle_state.rst
+++ b/docs/examples/vehicle_state.rst
@@ -18,10 +18,12 @@ The example can be run as described in :doc:`running_examples` (which in turn as
 the vehicle and DroneKit have been set up as described in :ref:`get-started`).
 
 If you're using a simulated vehicle remember to :ref:`disable arming checks <disable-arming-checks>` so 
-that the example can run. You can also 
-`add a virtual rangefinder <http://dev.ardupilot.com/wiki/using-sitl-for-ardupilot-testing/#adding_a_virtual_rangefinder>`_
-(otherwise the :py:attr:`Vehicle.rangefinder <dronekit.Vehicle.rangefinder>` attribute may return 
-values of ``None`` for the distance and voltage). 
+that the example can run. You can also:
+
+* `add a virtual rangefinder <http://dev.ardupilot.com/wiki/using-sitl-for-ardupilot-testing/#adding_a_virtual_rangefinder>`_
+  (otherwise the :py:attr:`Vehicle.rangefinder <dronekit.Vehicle.rangefinder>` attribute may return values of ``None`` for the distance and voltage).
+* `add a virtual gimbal <http://dev.ardupilot.com/wiki/using-sitl-for-ardupilot-testing/#adding_a_virtual_gimbal>`_
+  (otherwise the :py:attr:`Vehicle.gimbal <dronekit.Vehicle.gimbal>` attribute may return values of ``None`` for the yaw, pitch and roll). 
 
 In summary, after cloning the repository:
 
--- a/docs/examples/vehicle_state.rst
+++ b/docs/examples/vehicle_state.rst
@@ -64,7 +64,7 @@ On the command prompt you should see (something like):
      Attitude: Attitude:pitch=0.00294387154281,yaw=-0.11805768311,roll=0.00139428151306
      Velocity: [-0.03, 0.02, 0.0]
      GPS: GPSInfo:fix=3,num_sat=10
-     Mount status: [None, None, None]
+     Gimbal status: Gimbal: pitch=None, roll=None, yaw=None
      Battery: Battery:voltage=12.587,current=0.0,level=100
      EKF OK?: False
      Last Heartbeat: 0.769999980927
--- a/docs/guide/vehicle_state_and_parameters.rst
+++ b/docs/guide/vehicle_state_and_parameters.rst
@@ -27,7 +27,7 @@ Vehicle state information is exposed through vehicle *attributes*. DroneKit-Pyth
 :py:attr:`Vehicle.attitude <dronekit.Vehicle.attitude>`,
 :py:attr:`Vehicle.velocity <dronekit.Vehicle.velocity>`,
 :py:attr:`Vehicle.gps_0 <dronekit.Vehicle.gps_0>`,
-:py:attr:`Vehicle.mount_status <dronekit.Vehicle.mount_status>`,
+:py:attr:`Vehicle.gimbal <dronekit.Vehicle.gimbal>`,
 :py:attr:`Vehicle.battery <dronekit.Vehicle.battery>`,
 :py:attr:`Vehicle.rangefinder <dronekit.Vehicle.rangefinder>`,
 :py:attr:`Vehicle.ekf_ok <dronekit.Vehicle.ekf_ok>`,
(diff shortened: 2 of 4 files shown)


mrpollo:
looks good to me, are we waiting for @djnugent to include @hamishwillee suggestions?, I'm assuming we can close #497 now

djnugent:
Isn't this missing `@property gimbal()` for the  vehicle object? And this also doesn't deprecate mount_status.

hamishwillee:
Yes @djnugent is correct - missing the gimbal property on vehicle and deprecation of the mount_status. Also I need to add some updates to our other docs.

@mrpollo I'd quite like to know how this was verified/tested and confirmation of my questions. With respect to my suggestions we should minimally support the location types we have as inputs. Deciding to add a body/local frame is another question ....

I'm going to have a play with this now on SITL and will update "the basics".

hamishwillee:
Added a change to:
- add the gimbal property on the vehicle
- make Gimbal not be a HasObservable. This would be useful if we needed to individually observe roll, pitch, yaw, but I don't believe that is the case
- Update reference
- Changed target_gps method to target_location.

Did not:
- Update other docs/guide/examples (still needed)
- Add support for the different types of locations (IMO still needed)

I haven't yet been able to test on simulator due to https://github.com/diydrones/ardupilot/issues/3322 and https://github.com/dronekit/dronekit-sitl/issues/59. 

I did however test on Solo and had some problems. What I observed is that:
- We do get information from the gimbal, and we can observe it and see the value changes.
- When I try to set the position/orientation using ``rotate`` I don't see any movement of the real gimbal. I can see that the value returned by the callback for the set value does change, which makes me thing we may be (incorrectly) caching the value we set - @djnugent  is that possible?
- I attempted to set a ROI. Didn't want to wander too far from my computer so got inconclusive results. I believe that this probably was working ... at least to some extent. I THINK that this can only work in guided mode right - ie it won't work in stabilise mode.

Anyway, @djnugent can you verify that the rotate works OK on Solo for your tests, and ideally supply me the test code/script that verifies this. I'll keep trying to verify on simulator, but unless we look at that value caching problem it will always tell us we succeeded ... so not a "real" test.

djnugent:
Just tested on Solo and it works with this code
```
from dronekit import *
import time

vehicle = connect('0.0.0.0:14550', wait_ready=True)

print "pointing gimbal"
vehicle.gimbal.rotate(-45,0,0)
time.sleep(5)
print "roi"
loc = vehicle.location.global_relative_frame
vehicle.gimbal.target_gps(LocationGlobalRelative(loc.lat,loc.lon,loc.alt -100))
time.sleep(5)
print "releasing control"
vehicle.gimbal.release()
print vehicle.gimbal
time.sleep(0.5)
```

ROI is kind of spotty. After a ran this script a few times it became responsive. But the first few times were delayed. It was odd behavior but definitely an ardupilot thing.

djnugent:
I also confirmed gimbal read back is working. It does not give the actual IMU orientation of the gimbal but the last commanded orientation from the autopilot. This includes ROI, mavlink targeting, and, rc targeting. But it will not change if you poke your gimbal or rotate your copter.

hamishwillee:
HI @djnugent Thanks for that - roi stuff sounds similar to what I was seeing. Particularly useful to know that values are commanded values not measured - I will update docs appropriately. I wasn't seeing movement of the gimbal at all with rotate but I'll try that again. Did you see movement when you tried to set yaw and roll too?

hamishwillee:
@djnugent So I didn't realise the range of the camera pitch is 0 (straight ahead) to -90 (straight down). Pushing it to positive values not working all that well for me!

I tried this with yaw and roll values though, and couldn't see any movement of the gimbal. Did you test this?

djnugent:
Yaw controls solo's heading so I haven't tested that. And I don't think solo gimbal supports roll.

hamishwillee:
You did say that yaw was in global frame - makes sense now. I presume the roll command will just be ignored by the gimbal if not supported? (from what I can see the gimbal floats in the roll axis, so even if the vehicle rolled, the camera would remain level.

hamishwillee:
@djnugent et al. I've updated the documentation for all the methods and the example code. Can you please review the changes and confirm that how I think it works is actually what it does?

I made this so that setting an ROI can take either time of global location. Would be great to set ROIs locally, but too much to do for now. 

There is one more major change we should consider. According to @peterbarker Solo also sends out the GIMBAL_REPORT message, and we should be able to get more information from this - we should at least confirm what it is sending out. 

From a documentation perspective I will also need to add some example code on setting the gimbal, and this requires that I get simulation working. Still working on that (it would not be a blocker to this going out, but I'd hope we'd get the GIMBAL_REPORT looked at.

tcr3dr:
@hamishwillee GIMBAL_REPORT is [here](http://mavlink.org/messages/ardupilotmega#GIMBAL_REPORT):

1. Joint roll / elevation / azimuth (spherical coordinates)
1. Joint vx/vy/vz velocity
1. Joint x/y/z angle
1. Time since last update

Reading through the ardupilot source, it looks like this is too low-level to expose. It's pre-EKF, so the values are less reliable, and it's heartbeat-based from the MAVLink-enabled gimbal rather than from the autopilot. I would be inclined not to expose these in this PR.

So given that, can we merge and open a new issue for enabling the gimbal simulation? Several of us are interested in the simulation bits and testing this is pretty necessary, so I agree we want that soon.

hamishwillee:
@tcr3dr Yes. Essentially the GIMBAL_REPORT is a Solo only feature, so we would consider adding it to SoloVehicle, but not to generic DroneKit-Python. Simulation of gimbals is poor, but I'll add what information there is and post. This should be ready for this to be merged within the next hour.

I do feel that I personally have not tested it as much as I'd like. For example, if doing a yaw normally you can only do it AFTER sending a movement command (you can't just send a yaw during takeoff). Since setting gimbal yaw actually causes the vehicle yaw to change, I'd quite like confirmation that this isn't a limitation on setting gimbal behaviour.

hamishwillee:
@tcr3dr This is IMO ready to merge. Last minor update above points to new section on gimbal simulation I just put into dev wiki: http://dev.ardupilot.com/wiki/using-sitl-for-ardupilot-testing/#adding_a_virtual_gimbal

FYI, 
- this simulation is very poor (just about to create a defect report). Essentially it does allow you to start getting MOUNT_STATUS messages but our commands from dronekit (and mission planner) appear to be ignored. Specifically on Solo you send a yaw command then the vehicle yaws - but on simulator nothing happens. I think GIMBAL_REPORT is too high level, but it would be nice to know that the "intention" of the command had actually been followed :-0
- Gimbal simulation breaks Native windows/Cygwin and DK-SITL. @peterbarker has very kindly worked out a fix for vagrant-based SITL.

ANyway, good to merge!

tcr3dr:
@hamishwillee Loud and clear. @wsilva32 you might be interested in the above information on SITL's compatibility with gimbal right now.

hamishwillee:
Thank you. @wsilva32 FYI, these are some of the issues on the topic of simulation https://github.com/diydrones/ardupilot/issues/3335 , https://github.com/dronekit/dronekit-sitl/issues/59 , https://github.com/diydrones/ardupilot/issues/3322 , https://github.com/diydrones/ardupilot/pull/3325

PR #507, opened 2015-12-30
Hgw simulation update

hamishwillee:
This updates the SITL instructions to refer to DroneKit-Sitl as the main simulation option.

### DIFF
--- a/docs/develop/sitl_setup.rst
+++ b/docs/develop/sitl_setup.rst
@@ -4,103 +4,124 @@
 Setting up a Simulated Vehicle (SITL)
 =====================================
 
-The `SITL (software in the loop) <http://dev.ardupilot.com/wiki/simulation-2/sitl-simulator-software-in-the-loop/>`_ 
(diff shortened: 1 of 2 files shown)


hamishwillee:
@mrpollo This is probably OK to go in, though I just finished it and I will be reviewing it again tomorrow. 

Tomorrow I plan to see if the examples need to be updated in line with this. I am not quite sure how to do this as some examples could easily connect direct to the SITL TCP address while others would benefit from viewing using a GCS (and the setup instructions are a bit different). 
There are also comments like

>  "If you’re using a simulated vehicle, remember to disable arming checks"

This is not needed on DK-SITL.

I will probably suggest using MAVProxy port forwarding in all examples to keep things consistent and add a note in the simulation doc that we recommend doing this forwarding. Thoughts?

mrpollo:
Going to merge this in, we can review the examples on a new PR, better smaller PR's than 1 big monster!

looks great :+1:

hamishwillee:
Thank you - I'm more than happy to have it in :-)

Doesn't appear to have rebuilt on the site though?

mrpollo:
just pushed

PR #508, opened 2015-12-31
an example was using vehicle.gimbal without checking if it's valid fi…

davidbuzz:
…rst.

### DIFF
--- a/examples/vehicle_state/vehicle_state.py
+++ b/examples/vehicle_state/vehicle_state.py
@@ -31,7 +31,8 @@ print " Local Location: %s" % vehicle.location.local_frame
 print " Attitude: %s" % vehicle.attitude
 print " Velocity: %s" % vehicle.velocity
 print " GPS: %s" % vehicle.gps_0
-print " Gimbal status: %s" % vehicle.gimbal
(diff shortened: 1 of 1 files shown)


hamishwillee:
@davidbuzz Thanks for this. Traditionally we've assumed existence - ie the examples reflect the head revision of DK-Python. I'm not sure whether that is better, or checking as you have done.

peterbarker:
@davidbuzz Thanks for the PR, but I'm with Hamish on this one.

The examples are in the same repo as the code; I believe it is reasonable for them to reflect the code in the checkout with them, and that adding checks to all of the examples based on differences between older and newer versions would unnecessarily clutter the example code.

I'm curious to know how you came across this problem.  I'm guessing you had an install, but were running the examples from an updated git repo?

If you can make a strong use case, we're happy to reconsider :-)

Again, thanks.

davidbuzz:
FYI:  I did nothing special, I just didn't have a gimbal attached, and the code appeared to assume that a gimbal would always be attached.

hamishwillee:
All attributes return `None` if an attribute is not supported. From this example point of view there is no need to test for gimbal support.

PR #509, opened 2016-01-06
Versions and capability

squilter:
This creates two new classes called `version` and `capability`.  The information in these classes is parsed from the `AUTOPILOT_VERSION` message that is sent by the autopilot.  Unfortunately, ardupilot does not send this message automatically, it only sends it after it is requested.  The proper way to request it is with a do-command, like this:

```
#Request an AUTOPILOT_VERSION packet
capability_msg = self.message_factory.command_long_encode(0, 0, mavutil.mavlink.MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES, 0, 1, 0, 0, 0, 0, 0, 0)
self.send_mavlink(capability_msg)
```

I tried putting this into the vehicle.init() method, but it didn't work for some reason.  @tcr3dr do you know where is the best place to put this?

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -299,7 +299,7 @@ class Version(object):
             prefix += "Rover-"
 
         release_type="-dev"
-        if(release != None):
+        if(self.release != None):
             if(self.release == 255):
                 release_type = ""
             if(self.release > 192-1):
@@ -1953,6 +1953,10 @@ class Vehicle(HasObservers):
         if rate != None:
             self._master.mav.request_data_stream_send(0, 0, mavutil.mavlink.MAV_DATA_STREAM_ALL,
                                                       rate, 1)
+
+        #Request an AUTOPILOT_VERSION packet
+        capability_msg = self.message_factory.command_long_encode(0, 0, mavutil.mavlink.MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES, 0, 1, 0, 0, 0, 0, 0, 0)
+        self.send_mavlink(capability_msg)
 
         # Ensure initial parameter download has started.
         while True:
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -257,15 +257,24 @@ class Version(object):
     Version numbers.
 
     An object of this type is returned by :py:attr:`Vehicle.version`
-
     The version number can be read in a few different formats. To get it in a human-readable
     format, just print `vehicle.version`.  This might print somthing like "APM:Copter-3.3.2-rc4".
 
-    Or read the version in its four parts: major, minor, patch, release_type, by calling 
-    `vehicle.version.major`, `vehicle.version.minor`, `vehicle.version.patch` and `vehicle.version.release_type`.
-    Release_type is of the enum `FIRMWARE_VERSION_TYPE` which can be found in `mavutil.mavlink.FIRMWARE_VERSION_TYPE`
+    .. py:attribute:: major
 
-    To check if the version is a stable release, check `vehicle.version.number.is_stable()`
+    Major version number (integer).
+
+    .. py:attribute::minor
+
+    Minor version number (integer).
+
+    .. py:attribute:: patch
+
+    Patch version number (integer).
+
+    .. py:attribute:: release
+
+    Release type (integer). See the enum `FIRMWARE_VERSION_TYPE <http://mavlink.org/messages/common#MAV_AUTOPILOT_GENERIC>`_.
     """
     def __init__(self, raw_version, autopilot_type, vehicle_type):
         self.autopilot_type = autopilot_type
@@ -275,7 +284,7 @@ class Version(object):
             self.major = None
             self.minor = None
             self.patch = None
-            self.stable = None
+            self.release = None
         else:
             self.major   = raw_version >> 24 & 0xFF
             self.minor   = raw_version >> 16 & 0xFF
@@ -283,6 +292,10 @@ class Version(object):
             self.release = raw_version & 0xFF
 
     def is_stable(self):
+        """
+        Returns True if the autopilot reports that the current firmware is a stable
+        release, not a pre-release or development version.
+        """
         return self.release == 255
 
     def __str__(self):
@@ -314,13 +327,57 @@ class Capabilities:
     """
     The capabilities tells us what messages the autopilot is capable of interpreting.
 
-    To check if the autopilot support terrain handling, check if
-    `version.capabilities.terrain_handling is True`.
+    .. py:attribute:: mission_float
 
-    The other possible capabilities to check are `mission_float`, `param_float`,
-    `mission_int`, `command_int`, `param_union`, `fix_type`, ftp`, `set_attitude_target`,
-    `set_attitude_target_local_ned`, `set_altitude_target_global_int`, `actuator_target`,
-    `flight_termination`, `compass_calibration`.
+    Autopilot supports MISSION float message type. (Boolean)
+
+    .. py:attribute:: param_float
+
+    Autopilot supports the new param float message type. (Boolean)
+
+    .. py:attribute:: mission_int
+
+    Autopilot supports MISSION_INT scaled integer message type. (Boolean)
+
+    .. py:attribute:: command_int
+
+    Autopilot supports COMMAND_INT scaled integer message type. (Boolean)
+
+    .. py:attribute:: param_union 
+
+    Autopilot supports the new param union message type. (Boolean)
+
+    .. py:attribute:: ftp
+
+    Autopilot supports ftp for file transfers. (Boolean)
+
+    .. py:attribute:: set_attitude_target
+
+    Autopilot supports commanding attitude offboard. (Boolean)
+
+    .. py:attribute:: set_attitude_target_local_ned
+
+    Autopilot supports commanding position and velocity targets in local NED frame. (Boolean)
+
+    .. py:attribute:: set_attitude_target_global_int
+
+    Autopilot supports commanding position and velocity targets in global scaled integers. (Boolean)
+    
+    .. py:attribute:: terrain
+
+    Autopilot supports terrain protocol / data handling. (Boolean)
+
+    .. py:attribute:: set_actuator_target
+
+    Autopilot supports direct actuator control. (Boolean)
+
+    .. py:attribute:: flight_termination
+
+    Autopilot supports the flight termination command. (Boolean)
+
+    .. py:attribute:: compass_calibration
+
+    Autopilot supports onboard compass calibration. (Boolean)
     """
     def __init__(self, capabilities):
         self.mission_float                  = (((capabilities >> 0)  & 1) == 1)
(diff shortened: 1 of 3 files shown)


dronekit/__init__.py
```diff
@@ -254,6 +252,73 @@ def __init__(self, distance, voltage):
...
+class Version(object):
+    """
+    The version number in a few different formats. To get it in a human-readable
+    format, just print `vehicle.version.number`. Or read the version in its four
+    parts: major, minor, patch, release_type, by calling `vehicle.version.number.major`, etc.
+
+    To check if the version is a stable release, check `vehicle.version.number.is_stable()`
+    """
+    def __init__(self, raw_version, autopilot_type, vehicle_type):
+        self.autopilot_type = autopilot_type
```
hamishwillee: Each of the attributes should be separately documented properties.

dronekit/__init__.py
```diff
@@ -254,6 +252,73 @@ def __init__(self, distance, voltage):
...
 
+class Version(object):
+    """
+    The version number in a few different formats. To get it in a human-readable
+    format, just print `vehicle.version.number`. Or read the version in its four
+    parts: major, minor, patch, release_type, by calling `vehicle.version.number.major`, etc.
+
+    To check if the version is a stable release, check `vehicle.version.number.is_stable()`
+    """
+    def __init__(self, raw_version, autopilot_type, vehicle_type):
```
hamishwillee: Not every version will send the message. We use the convention that `None` is used by default for attribute values on object creation, and these are then populated with values on the first message received. That way it is obvious to users of the API when the attribute isn't supported.

dronekit/__init__.py
```diff
@@ -254,6 +252,73 @@ def __init__(self, distance, voltage):
     def __str__(self):
         return "Rangefinder: distance={}, voltage={}".format(self.distance, self.voltage)
 
+class Version(object):
+    """
+    The version number in a few different formats. To get it in a human-readable
+    format, just print `vehicle.version.number`. Or read the version in its four
+    parts: major, minor, patch, release_type, by calling `vehicle.version.number.major`, etc.
```
hamishwillee: There does not appear to be a number attribute, so would this not fail?

dronekit/__init__.py
```diff
@@ -889,6 +954,14 @@ def listener(self, name, m):
             self._mount_yaw = m.pointing_c / 100
             self.notify_attribute_listeners('mount', self.mount_status)
 
+        self._capabilities = 0
+        self._raw_version = 0
+
+        @self.on_message('AUTOPILOT_VERSION')
```
hamishwillee: Having this here works, but it is cleaner to include the `on_message` in the class that gets populated IFF this is likely to be the only listener for the message.

dronekit/__init__.py
```diff
@@ -254,6 +252,73 @@ def __init__(self, distance, voltage):
...
+        return prefix + "-%s.%s.%s" % (self.major, self.minor, self.patch) + release_type
+
+class Capabilities:
+    """
+    The capabilities tells us what messages the autopilot is capable of interpreting.
+    For example, to check if the autopilot supports parachutes, check if `version.capabilities.parachute is True`.
+    Note: This doesn't necessarily mean that a parachute is connected and configured. It only means that the
+    autopilot knows what a parachute is.
+    """
+    def __init__(self, capabilities):
```
hamishwillee: As above, as attributes with separate documentation is the standard approach.

dronekit/__init__.py
```diff
@@ -1392,6 +1469,20 @@ def velocity(self):
         return [self._vx, self._vy, self._vz]
 
     @property
+    def version(self):
+        """
+        The autopilot version in a Version object.
```
hamishwillee: Replace `Version` with
```
:py:class:`Version`
```
To create a link to your version class

dronekit/__init__.py
```diff
@@ -1392,6 +1469,20 @@ def velocity(self):
...
+        """
+        return Version(self._raw_version, self._autopilot_type, self._vehicle_type)
+
+    @property
+    def capabilities(self):
+        """
+        The capabilities of the autopilot in a Capabilities object.
```
hamishwillee: Replace `Capabilities` in this sentence with
```
:py:class:`Capabilities`
```
To create a link to your Capabilities class

dronekit/__init__.py
```diff
@@ -47,8 +47,6 @@
 import copy
 import collections
 from pymavlink.dialects.v10 import ardupilotmega
```
hamishwillee: @tcr3dr @mrpollo  Is this change reasonable?

hamishwillee:
@squilter Thanks for this! I've added some comments ...

I don't "know" the answer to:

>  tried putting this into the vehicle.init() method, but it didn't work for some reason. @tcr3dr do you know where is the best place to put this?

But I'd suggest trying in [Vehicle.initialize](https://github.com/squilter/dronekit-python/blob/versions_and_capability/dronekit/__init__.py#L1936] since we set up the data stream there.

Could you create a [test class for this here](https://github.com/squilter/dronekit-python/tree/versions_and_capability/dronekit/test/sitl)?

For extra kudos, it would also be cool if you could update [vehicle state example](https://github.com/squilter/dronekit-python/blob/versions_and_capability/examples/vehicle_state/vehicle_state.py). If you don't, then I will when I actually test this in anger.

dronekit/__init__.py
squilter: Maybe I'm understanding wrong, but if the `on_message` is in the `Capabilities` object, then the listener won't start running until after the object is instantiated (i.e. when somebody tries check a capability).  Since the autopilot will only send one `AUTOPILOT_VERSION` (rather than sending it multiple times per second), we need that listener to be running from the start to catch it, not after the user requests the data.

squilter:
All works on my computer.  Not sure why the test is failing.  Maybe it's because of the version that sitl is running?  If sitl is running an older version, then it would send lots of 0's in the `AUTOPILOT_VERSION` message, which would lead to this failing.

hamishwillee:
@squilter I've reviewed and added some notes. Thank you for adding some documentation. I would prefer that all public items were exposed as attributes (or functions) and individually documented. This is primarily for consistency, because even if it results in some more code, it ensures all the code looks the same and is documented in the same way. That is comforting for developers!

@mrpollo and @tcr3dr might take a different view. @mrpollo Can you do a technical review on the contribution? I have not done "full testing" yet.

squilter:
I'm still not exactly sure how you want the documentation to look.

hamishwillee:
@squilter What you should do in this case is copy the model of [SystemStatus](https://github.com/squilter/dronekit-python/blob/versions_and_capability/dronekit/__init__.py#L406)

So for each public item you'd create an attribute in your description:
```
    .. py:attribute:: major

        Major version number (integer).
```

This allows me to link to the variable.  You should also create documentation in your public methods - again creating linkable objects: eg.
```
def is_stable(self):
   """
    Your description here.
    """
   return self.release == 255
```

FYI, you were absolutely right be confused. We have been inconsistent - using `param` in many cases rather than attribute definition. I will change these in a future update.

YOu also don't need to create variables as python attributes as I suggested in previous notes, since we haven't mostly done this for helper classes. Apologies for the incorrect suggestion.

PPS I still quite like having the attribute observers in the helper class init. It just feels cleaner to me.

squilter:
@tcr3dr I think the only thing left to do now is to fix the tests.  @tcr3dr what version of ArduCopter is running in the sitl tests?  If it's older than 3.3, that would explain why it's failing.

mrpollo:
its failing on your test @squilter, it's fetching latest copter from our hosting, right now it's: "`APM:Copter V3.3`"

squilter:
Okay I found it.  The problem is in ArduCopter.  I'll talk to Randy and try to fix it, but the problem already exists in 3.3, so we need to figure out how to proceed here.

After sending a `MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES`, ArduCopter responds with a `AUTOPILOT_VERSION` packet, which includes a version number (e.g. 3.4.0-dev), and a bitmask of supported capabilities.  The problem is that if this packet is requested in the first few seconds after the vehicle boots, it will send back a packet with `0`'s in the capability bitmask (version field is fine).  

So it worked fine for me because I started SITL, then started dronekit.  If it happens the other way around, the packet that is received in response will have an empty `capabilities` field, causing the test to fail.

squilter:
Fix for arducopter is at diydrones/ardupilot#3485

squilter:
The problem only exists in Copter 3.3.  It will be fixed in the next copter, and isn't an issue in other vehicles.  It only occurs if dronekit connects to the vehicle before the vehicle is powered on.  I think we should remove this particular check from the test for now, and add it again after the next version is published.

squilter:
.

dronekit/__init__.py
```diff
@@ -254,6 +252,148 @@ def __init__(self, distance, voltage):
...
+    """
+    Version numbers.
+
+    An object of this type is returned by :py:attr:`Vehicle.version`
+    The version number can be read in a few different formats. To get it in a human-readable
+    format, just print `vehicle.version`.  This might print somthing like "APM:Copter-3.3.2-rc4".
+
+    .. py:attribute:: major
+
+    Major version number (integer).
```
hamishwillee: Indent 4 spaces (this is how sphinx works out that it is processing a new block). In fact for any new directive you have to have a new line before the directive, a line after the directive, and then the content of the directive block must be indented by 4 spaces. So for example a note would look like this:

```
some text.

.. note::

    The note content.
    This is more note content.
```

dronekit/__init__.py
```diff
@@ -254,6 +252,148 @@ def __init__(self, distance, voltage):
...
+
+    .. py:attribute::minor
+
+    Minor version number (integer).
```
hamishwillee: Indent 4 spaces.

dronekit/__init__.py
```diff
@@ -254,6 +252,148 @@ def __init__(self, distance, voltage):
...
+
+    .. py:attribute:: patch
+
+    Patch version number (integer).
```
hamishwillee: Indent 4 spaces.

dronekit/__init__.py
```diff
@@ -254,6 +252,148 @@ def __init__(self, distance, voltage):
...
+
+    .. py:attribute:: release
+
+    Release type (integer). See the enum `FIRMWARE_VERSION_TYPE <http://mavlink.org/messages/common#MAV_AUTOPILOT_GENERIC>`_.
```
hamishwillee: Indent 4 spaces.

dronekit/__init__.py
```diff
@@ -254,6 +252,148 @@ def __init__(self, distance, voltage):
...
+                release_type = "-alpha" + str(self.release-(192-1))
+        return prefix + "%s.%s.%s" % (self.major, self.minor, self.patch) + release_type
+
+class Capabilities:
+    """
+    The capabilities tells us what messages the autopilot is capable of interpreting.
+
+    .. py:attribute:: mission_float
+
+    Autopilot supports MISSION float message type. (Boolean)
```
hamishwillee: Indent 4 spaces.

hamishwillee:
@squilter The indentation isn't quite right for the attribute blocks (I've marked up a few) and I would prefer that the "attribute type" was inside the sentence as shown in [example above](https://github.com/dronekit/dronekit-python/pull/509#discussion_r50211206). The content and approach is great though - thanks very much!

@mrpollo What do you think of @squilter comments - https://github.com/dronekit/dronekit-python/pull/509#issuecomment-170630990
Can we wait or re-request the capabilities if we get a zero-filled bitmask?

peterbarker:
This was rebased against master and changed around a bit as part of #545

PR #513, opened 2016-01-12
Fix mount status to return float degrees, not int degrees

njoubert:
Currently MOUNT_STATUS returns integer degrees. But the MAVLINK message sends it as integer centi-degrees, so we're losing two orders of magnitude in precision. This fix changes MOUNT_STATUS to return float values.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -884,9 +884,9 @@ class Vehicle(HasObservers):
 
         @self.on_message('MOUNT_STATUS')
         def listener(self, name, m):
-            self._mount_pitch = m.pointing_a / 100
-            self._mount_roll = m.pointing_b / 100
-            self._mount_yaw = m.pointing_c / 100
+            self._mount_pitch = m.pointing_a / 100.0
+            self._mount_roll = m.pointing_b / 100.0
+            self._mount_yaw = m.pointing_c / 100.0
             self.notify_attribute_listeners('mount', self.mount_status)
 
         # gimbal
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @njoubert . @mrpollo this looks good to me. What do we need to do to get it merged?

mrpollo:
I apologize, our tests seem to be broken, let me fix those before we merge our outstanding PR's, @njoubert thanks for taking the time to add this in, would it be too much to ask for a regression test?

njoubert:
@mrpollo sure i can add something, but how do you guys do regression tests?

mrpollo:
just standard tests, check here for examples:

- [standalone unit tests](https://github.com/dronekit/dronekit-python/tree/master/dronekit/test/unit)
- [unit tests that run on a SITL environment](https://github.com/dronekit/dronekit-python/tree/master/dronekit/test/sitl)

pick what you think fits this test best

hamishwillee:
@njoubert @mrpollo I think unit tests, as "somewhat" documented [here](http://python.dronekit.io/contributing/contributions_api.html#unit-tests). The docs are a little thin because I don't really understand "mock". 

That said, unit tests make sense if you're only trying to verify that the API doesn't fall over if you push in a float value.

peterbarker:
Closing in favour of https://github.com/dronekit/dronekit-python/pull/541

Thanks @njoubert - I've simply rebased this and tested it works - which it did, of course :-)

PR #530, opened 2016-02-03
Update contribution process docs

hamishwillee:
Now specifies our real contribution approach - branch-off-main-repo, rather than suggesting people fork repo.


peterbarker:
hamishwillee
1:34 PM Well [darn]. Then suggest you don't merge this. Close with reason "we can't support this safely"

PR #533, opened 2016-02-05
Hgw performance example

hamishwillee:
This updates the performance example, getting rid of all the "rubbish" and updating it to DKPY 2.0.0. It also adds example documentation. It has been tested against SITL and Solo.

The example will use DK-SITL if no connection string is specified. This is the model I hope to roll out to the other examples.


peterbarker:
This PR removes performance test results which may answer the question, "is this result good or not?".  The results themselves are hideously out-of-date, 'though.

hamishwillee:
Thanks. I added #535 to track need to add new results.

PR #539, opened 2016-02-08
Make drone delivery example use DK-SITL simulator

hamishwillee:
Example now automatically uses DK-SITL if no connection string is specified. I am doing this for all examples (slowly). I also updated the docs (which were incorrect in several place anyway).


peterbarker:
LGTM.  Pity travis is wedged - and I have no access to retry.

PR #540, opened 2016-02-08
Update best_practice.rst

DanielHabib:
# Summary
I found duplication in a try/catch so I removed it.

#### Question
I used `API Error` as the comment above the error that is thrown since `dronekit.APIException` is fairly vague and not much can be deduced from that. 
Would it be preferred for this comment to be more descriptive?


peterbarker:
Thanks for the patch!

I rebased this on master as part of #544 and merged it.

The comment could have been more descriptive; it doesn't really add anything past what the exception name itself supplies :-)

PR #541, opened 2016-02-08
Fix mount status to return float degrees, not int degrees

peterbarker:
Rebased and tested this: https://github.com/dronekit/dronekit-python/pull/513

I used a real quad for testing this; mount simulation needs work.


hamishwillee:
Looks good to me. Merged.

PR #545, opened 2016-02-09
Versions and capability

peterbarker:
Rebased against master, made tests work against Copter3.3

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -254,11 +254,14 @@ class Rangefinder(object):
 
 class Version(object):
     """
-    Version numbers.
+    Autopilot version and type.
 
-    An object of this type is returned by :py:attr:`Vehicle.version`
+    An object of this type is returned by :py:attr:`Vehicle.version`.
+    
     The version number can be read in a few different formats. To get it in a human-readable
-    format, just print `vehicle.version`.  This might print somthing like "APM:Copter-3.3.2-rc4".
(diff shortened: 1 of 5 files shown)


dronekit/__init__.py
```diff
@@ -1845,6 +2018,12 @@ def initialize(self, rate=4, heartbeat_timeout=30):
             self._master.mav.request_data_stream_send(0, 0, mavutil.mavlink.MAV_DATA_STREAM_ALL,
                                                       rate, 1)
 
+        #Request an AUTOPILOT_VERSION packet
+        def send_capa_request(vehicle, name, m):
```
hamishwillee: I prefer full descriptive names. We're in Python. `send_capabilities_request` or even better `request_capabilities`. Does this need to be public or can it be hidden?

hamishwillee:
@peterbarker Is there any meaningful default printing we can/should do of Version.capabilities?

hamishwillee:
@peterbarker I'm getting an error trying to use `vehicle.capabilities.set_attitude_target_global_int`

```
print "   Supports set position + velocity targets in global scaled integers: %s" % vehicle.capabilities.set_attitude_target_global_int

AttributeError: Capabilities instance has no attribute 'set_attitude_target_global_int'
```
Thoughts?

hamishwillee:
@peterbarker The version.release is also just an integer. Can we improve what we provide here?

PR #554, opened 2016-02-10
Increase timeouts, factor out a wait_for call

peterbarker:


hamishwillee:
This looks good for me, and the tests run fine locally. I wasn't able to trigger a timeout to test it more broadly.

PR #559, opened 2016-02-11
Rejig flight replay example to use a telemetry log

peterbarker:


examples/flight_replay/flight_replay.py
```diff
@@ -7,122 +7,178 @@
...
-    """Download a public mission from droneshare (as JSON)"""
-    message_url="%s/api/v1/mission/%s/messages.json?max_freq=%s&api_key=%s" % (args.api_server, mission_id, max_freq, args.api_key)
-    f = urllib.urlopen(message_url)
-    j = json.load(f, object_hook=_decode_dict)
-    f.close()
-    return j
-
-def replay_mission(payload):
-    """Given mission JSON, set a series of wpts approximating the previous flight"""
+def start_default_sitl(lat=-35.363261, lon=149.165230):
```
hamishwillee: Lets pass these in as argument to the argparser, or infer them from the first recorded point in the log.

examples/flight_replay/flight_replay.py
```diff
@@ -7,122 +7,178 @@
...
+cmds = vehicle.commands
+cmds.clear()
+for i in xrange(0, len(messages)):
+    pt = messages[i]
+    print "Point: %d %d\n" % (pt.lat, pt.lon,)
+    lat = pt.lat
+    lon = pt.lon
+    # To prevent accidents we don't trust the altitude in the original flight, instead
+    # we just put in a conservative cruising altitude.
+    altitude = 30.0/100.0
```
hamishwillee: Is height 0.3 m reasonable. Previously we had at 30m. Bug?

examples/flight_replay/flight_replay.py
```diff
@@ -7,122 +7,178 @@
...
+vehicle.commands.next=0
+
+# Set mode to AUTO to start mission:
+vehicle.mode = VehicleMode("AUTO")
+
+while (vehicle.mode.name != "AUTO"):
+    time.sleep(0.1)
+
+print("Raising throttle to trigger mission start")
+vehicle.channels.overrides = {'3':1500}
```
hamishwillee: Find another way. We don't recommend or show channel overrides. There is a mission start command you can use to start from ground in a mission (if the mission has a takeoff command) but that is only present in copter 3.3. Might be better here to use an explicit takeoff method just like all the other examples.

hamishwillee:
@peterbarker Looks pretty good though the build failed. There are a few issues/comments above. When you're done, I would update the docs.

examples/flight_replay/flight_replay.py
peterbarker: Gah!  I set it all up in the right order to pull the start location out of the first message - and then forgot to do that!

examples/flight_replay/flight_replay.py
peterbarker: That's actually the way you do it on a real copter!

I thought it would be good to show some variety as to how you can do things.

You disagree?

hamishwillee: Yes, I disagree.

The "right way to do this" according to Randy is to use the takeoff command. We never recommend rc overrides except for simulation, and in fact I had to fight pretty hard to keep them in DK for version 2.0

In a real copter you take off then switch to auto -

peterbarker: No, I don't :-)

Where by, "real helicopter" I mean my quads, not an *actual* helicopter :-)

hamishwillee: Great. You can just change the standard takeoff code.

examples/flight_replay/flight_replay.py
```diff
@@ -7,122 +7,180 @@
...
+
+print("Arming vehicle")
+vehicle.armed = True
+
+while not vehicle.armed:      
+    print "Waiting for arming..."
+    time.sleep(1)
+
+print("Taking off")
+vehicle.simple_takeoff(30.0)
```
hamishwillee: This isn't "good code" since the command is asyncrhonous and if another command comes in it will be "replaced". It probably works because the time taken before someone is usually ready to start sending other commands is quite long ... so your discretion if it needs to change.

examples/flight_replay/flight_replay.py
hamishwillee: Still hard coded?

hamishwillee:
@peterbarker I've updated this so it now only keeps points that are 3m from the previous kept point, and only the first 99 of those points:
- This gives *much better* results than just taking evenly spread values across the range because the actual curves are retained. I don't think we're losing much detail because GPS doesn't have much greater resolution than 3m anyway. The cost is a slightly more complicated code (not ideal in an example) and we're only displaying a part of the mission.
- We could actually store all 277 points using this strategy, but I want code that can run safely whatever TLOG someone might throw at it. 

This also replaces your takeoff code with a "near standard" takeoff function. I say near standard because I kept your debug output (it is better than the output in simple_goto.py). The code is safer than your version because it checks that the vehicle is initialised and can arm (using is_armable). I also prefer a function because this will be tidier to update when we improve the API.

The docs have also been updated!

Ready for your review now.

PR #565, opened 2016-02-17
Check if mode_mapping has items before iteration

sanderux:
Simple sanity check to add basic support for PX4

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1161,7 +1161,8 @@ class Vehicle(HasObservers):
         def listener(self, name, m):
             self._armed = (m.base_mode & mavutil.mavlink.MAV_MODE_FLAG_SAFETY_ARMED) != 0
             self.notify_attribute_listeners('armed', self.armed, cache=True)
-            self._flightmode = {v: k for k, v in self._master.mode_mapping().items()}[m.custom_mode]
+            if self._master.mode_mapping() != None:
+                self._flightmode = {v: k for k, v in self._master.mode_mapping().items()}[m.custom_mode]
             self.notify_attribute_listeners('mode', self.mode, cache=True)
             self._system_status = m.system_status
             self.notify_attribute_listeners('system_status', self.system_status, cache=True)


LorenzMeier:
@tcr3dr @liamstask Would be nice to get this in rather fast!

mrpollo:
this looks good to me, would be nice if we could figure out how to add regression tests for this @sanderux, any ideas?

dronekit/__init__.py
```diff
@@ -1161,7 +1161,8 @@ def listener(self, name, m):
         def listener(self, name, m):
             self._armed = (m.base_mode & mavutil.mavlink.MAV_MODE_FLAG_SAFETY_ARMED) != 0
             self.notify_attribute_listeners('armed', self.armed, cache=True)
-            self._flightmode = {v: k for k, v in self._master.mode_mapping().items()}[m.custom_mode]
+            if self._master.mode_mapping() != None:
```
mrpollo: (just an FYI for anyone else looking at this) `None` here makes sense since upstream pymavlink `mode_mapping` is expected to return `None` https://github.com/mavlink/mavlink/blob/master/pymavlink/mavutil.py#L481-L502

mrpollo:
I'm merging this, appveyor chocked on the webhook call, tests are good

sanderux:
I guess you could perform tests by setting your SITL to identify as MAV_AUTOPILOT_INVALID

hamishwillee:
I guess it's time to run up a PX4 simulator and run our tests and example code. @sanderux What IP address is the simulator exposed on? It doesn't say here: http://dev.px4.io/simulation-sitl.html

LorenzMeier:
14550 GCS and 14540 for a secondary link:
https://github.com/PX4/Firmware/blob/master/posix-configs/SITL/init/rcS_jmavsim_iris#L58-L59

PX4 SITL can run headless, but we're just getting into it now in conjunction with Travis / Jenkins (outside of our ROS CI system which already runs on Jenkins headless): https://github.com/PX4/Firmware/pull/3790

http://sitl01.dronetest.io

hamishwillee:
Thanks @LorenzMeier . Any thoughts on whether it could be nicely integrated into DroneKit-Sitl :-)

PR #570, opened 2016-02-18
Splits outbound messages into its own thread.

tcr3dr:
We maintain a separate thread for outbound messages so that the main thread as well as the reading thread can send messages.

Right now outbound messages are sent in lockstep with reading messages; this actually creates huge problems with high-rate telemetry (on Solo proper). Splitting it into a concurrent thread passes tests but also improves output rate greatly.

Also, we add use_native.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -2705,7 +2705,8 @@ def connect(ip,
             rate=4,
             baud=115200,
             heartbeat_timeout=30,
-            source_system=255):
+            source_system=255,
+            use_native=False):
     """
     Returns a :py:class:`Vehicle` object connected to the address specified by string parameter ``ip``. 
     Connection string parameters (``ip``) for different targets are listed in the :ref:`getting started guide <get_started_connecting>`.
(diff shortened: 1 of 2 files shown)


tcr3dr:
This looks like a bump for 2.2.0

tcr3dr:
Ugh I apologize for the whitespace mess. Let me fix that.

tcr3dr:
Commits are now fixed.

tcr3dr:
Hold off on this for a moment—having issues on a real drone.

tcr3dr:
Ready for merge.

NOTE: MAVNative parses messages as generic MAVLINK_message objects, not as a specific MAVLink_message_class_type. This is what I worked around in @ba7f1c5

hamishwillee:
@tcr3dr When should you use/not use ``use_native`` (Use precompiled MAVLink parser). What difference does this make?

peterbarker:
This looks OK with the following caveats:
 - we're effectively throwing away whatever is causing the pack problems; we should probably understand why the pack fails
 - we're catching EVERY exception that occurs, both in the pack and the queueing; should probably tighten that up a bit

peterbarker:
I've squashed some commits and forced it up to try to make appveyor happy.  No luck so far.

mrpollo:
appveyor is not playing nice on this one, I'm afraid the limited resources available for the jobs to run is causing timeouts when running 2 instances of SITL

mrpollo:
appveyor is a mystery on its own that we can resolve here #574

this changeset looks good to me :+1:  as it stands

Issue #574, opened 2016-02-19
appveyor environment is causing timeout issues


mrpollo:
https://ci.appveyor.com/project/3drobotics/dronekit-python/build/1.0.352

tcr3dr:
This is probably super specific to the MAVLink forwarding feature we have. I imagine all other tests are passing on Windows. A developer with Windows would be able to debug that MAVLink test fairly straightforwardly.

mrpollo:
I disabled the Appveyor CI until we figure out how to fix it

pietrodn:
AppVeyor [works](https://ci.appveyor.com/project/pietrodn/dronekit-python/builds/20386105) on the latest master and can be re-enabled. See #878.

pietrodn:
AppVeyor builds are now enabled, running and successful. This issue can be closed. 🥳

peterbarker:
Closed, thanks!

PR #575, opened 2016-02-19
bump release to 2.2.0

mrpollo:
- Splits outbound messages into its own thread.
- remove of capabilities request on HEARTBEAT listener
- check if mode_mapping has items before iteration


mrpollo:
this looks good to me pending #574 but thats going to have to wait for 2.2.1

PR #577, opened 2016-02-19
Validate existence of custom_mode before setting	

mrpollo:
rebase of #569


tcr3dr:
This lgtm, but you might want to address this:

Now that we set `self._flightmode = 'AUTO'` by default, there is no circumstance under which `self._flightmode` will be None. But we check for this in several places. Should you remove those checks?

mrpollo:
I'm going to try and cleanup some of this logic I'll check for any `_flightmode` checks as well, we should proceed with merging this so we can unblock the px4 team

PR #587, opened 2016-02-24
Swallow message errors and continue.

tcr3dr:
Because a failed "*" handler can kill all subsequent message handlers, it becomes very hard to track down where an error is being found.

This patch handles errors consistently with how they are handled at a higher level (in `mavlink_thread_in`). A more robust error handling mechanism (with configurable logging) is desired, but this patch is more essential for now.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1464,9 +1464,21 @@ class Vehicle(HasObservers):
 
     def notify_message_listeners(self, name, msg):
         for fn in self._message_listeners.get(name, []):
-            fn(self, name, msg)
+            try:
+                fn(self, name, msg)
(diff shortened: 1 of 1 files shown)


hamishwillee:
@tcr3dr Tests failed - no idea how to unblock appveyor. This looks sane. Any way we could print out the listener function name if it exists as well - would make debugging just a little easier.

hamishwillee:
Also this is labelled WIP, so won't do anything until you say :-)

peterbarker:
:+1: LGTM

tcr3dr:
@peterbarker This is no longer a WIP, since it's about done. We do not need it in production, but it can still land if anyone is interested.

PR #594, opened 2016-02-26
time warp!

### LINKED ISSUE #592: time.time is not monotonic
we are facing some timeout issues when the OS clock jumps in time due to the clock being adjusted, we need to move to a monotonic time implementation for our timeouts and heartbeat checks

mrpollo:
use the monotonic clock for timeouts all over the API

fixes #592

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ psutil>=3.0.0
 mock>=1.3.0
 six>=1.9.0
 dronekit-sitl>=3.0,<=3.99999
+monotonic==0.6
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,4 @@ psutil>=3.0.0
 mock>=1.3.0
 six>=1.9.0
 dronekit-sitl>=3.0,<=3.99999
-monotonic==0.6
+monotonic<1.0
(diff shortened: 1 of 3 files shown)


hamishwillee:
@peterbarker YOu want to check this one out?

requirements.txt
```diff
@@ -6,3 +6,4 @@ psutil>=3.0.0
 mock>=1.3.0
 six>=1.9.0
 dronekit-sitl>=3.0,<=3.99999
+monotonic==0.6
```
peterbarker: Any reason this is strictly version 0.6?

setup.py
```diff
@@ -12,6 +12,7 @@
       author='3D Robotics',
       install_requires=[
           'pymavlink>=1.1.62',
+          'monotonic==0.6'
```
peterbarker: Any reason this is strictly version 0.6?

peterbarker:
:+1: LGTM with the exception of that version check on monotonic.

requirements.txt
mrpollo: no particular reason, I'll send a new commit for `<1.0`

setup.py
mrpollo: no particular reason, I'll send a new commit for `<1.0`

mrpollo:
thanks @peterbarker, we'll get this merged since its blocking a Solo release

PR #596, opened 2016-02-28
Fix ports in example documents

dtaniwaki:
Hi there,

Could you update [the example documents](http://python.dronekit.io/examples/vehicle_state.html) since the default port is not `127.0.0.1: 14550` anymore and it's confusing to try the examples?
I also fixed the wrong value in the current docs just in case.


hamishwillee:
Thanks very much @dtaniwaki. Looks good to me.

PR #604, opened 2016-03-11
Support Python 3 as well as Python 2

Rosuav:
This is the very beginnings of support. Hopefully other people can start testing and finding where the issues are.

This is enough to make `import dronekit` work under Py3.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -40,7 +40,8 @@ import platform
 import re
 from dronekit.util import errprinter
 from pymavlink import mavutil, mavwp
-from Queue import Queue, Empty
+try: from Queue import Queue, Empty # Python 2
+except ImportError: from queue import Queue, Empty # Python 3
 from threading import Thread
 import types
 import threading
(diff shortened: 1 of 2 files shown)


tcr3dr:
Thanks for the PR @Rosuav!

To whomever will merge this, I strongly should suggest we support Python 3 using the `six` library, but this is definitely an adequate fix in my book.

Rosuav:
Yeah, this was just a quick fix, nothing more. I haven't run the tests or anything. My goal is for this PR to be a focal point for people who actually use the software :) My experience with dronekit is limited to helping one of my students to install and use it; I got him to hack his copy to "from queue import", and then made the most minimal change possible for the PR.

mrpollo:
@Rosuav would you be able to make changes to your branch to reflect on @tcr3dr comments?, I think they are valid we should use the six library

Rosuav:
Not sure how much of `six` is needed here, and I actually don't have any experience with dronekit. (Nor much experience with six either, to be honest; I generally minimize dependencies, which means a few try/except imports rather than a third-party module.) It would be best of this is done by people who can actually play around with the module and use it for real.

mrpollo:
closed in favor of #655

PR #605, opened 2016-03-13
MAV Type for rover was incorrect.

adricl:
When connecting to AMPRovers this failed due to the incorrect type, This is now corrected


hamishwillee:
@adricl Thanks very much - you're absolutely right. Merging.

Issue #610, opened 2016-04-03
Exception in message handler for heartbeat


dgorissen:
I found 3 threads about this but no authorative answer/solution:

https://groups.google.com/forum/#!topic/drones-discuss/tQQjc0_5Eng
https://discuss.dronekit.io/t/issue-with-vehicle-state-py-and-rover/307
https://discuss.dronekit.io/t/missionplanner-and-dronekit-floods-console-with-errors/287

Im running dronekit 2.4.0 on an odroid UX4, connected with an FTDI cable to the Pixhawk Telem 2 port (dronekit is reading straight from /dev/ttyUSB0). Pixhawk is running APM copter 3.3.3 and the 3DR Radio is on Telem 1.  Serial 4/5 is also in use (Taranis telemetry).

I notice the following messages being spammed to std err/out on the odroid. Otherwise it does seem to be working properly (both reading the data via dronekit and apm planner 2.0).

```
...
argument of type 'NoneType' is not iterable
Exception in message handler for HEARTBEAT
argument of type 'NoneType' is not iterable
Exception in message handler for HEARTBEAT
argument of type 'NoneType' is not iterable
Exception in message handler for HEARTBEAT
...
```

It seems the issue is something to do with the telem1 & telem2 streams interfering with each other somehow? Some guidance as to what the correct usage pattern is would be appreciated. Or confirm if this is a bug.

PureRockets:
I am having the same error, however it is critical for me.  I'm using an ErleBrain2 running APM. It always returns Not Armable even when it's armed. When I try to run vehicle_state, I get several of these heartbeat exceptions, then it crashes (same as reported in 307)

peterbarker:
Attempting to iterate over NoneType is clearly a bug.  We'll need more debug to try to track this.

Could you add the following lines:

                import traceback
                traceback.print_exc()


into mavlink.py, just under this line:
                                errprinter('>>> ' + str(e))

please?

dgorissen:
It was actually in a slightly different place that the edit was needed but none the less:

```
>>> Exception in message handler for HEARTBEAT
>>> argument of type 'NoneType' is not iterable
Traceback (most recent call last):
  File "/home/dgorissen/.virtualenvs/orangspy/local/lib/python2.7/site-packages/dronekit/mavlink.py", line 150, in mavlink_thread_in
    fn(self, msg)
[...]

PR #611, opened 2016-04-04
#610: Wrap iterable call in try..except to deal with unrecognised vehicle codes

fnoop:
Mode code isn't always in the mode mappings - for example mode 0 that GCS seem to use.  Handle the exception and return a false which seems to be logically correct return in this case.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1521,7 +1521,10 @@ class Vehicle(HasObservers):
         return mavutil.mode_mapping_bynumber(self._vehicle_type)
 
     def _is_mode_available(self, mode_code):
-        return mode_code in self._mode_mapping_bynumber
+        try:
+            return mode_code in self._mode_mapping_bynumber
+        except:
+            return False
 
     #
     # Operations to support the standard API.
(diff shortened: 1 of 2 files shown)


mrpollo:
excellent contribution, would you have time to add integration tests?

fnoop:
@mrpollo - Hi, I'm brand new to git development so haven't got a clue I'm afraid.  But I'd like to learn and give it a go, if you could point me in the right direction?  Even just an existing example I can look at?

mrpollo:
sure thing @fnoop 

take a look at the tests we have here https://github.com/dronekit/dronekit-python/tree/master/dronekit/test/sitl

fnoop:
Note: The problem isn't that the mode code isn't in the mode mappings, it's that the mode mappings aren't created for all vehicle types, so _mode_mapping_bynumber() returns None for certain or unrecognised vehicle types (eg GCS).  Thus _is_mode_available() then fails doing a lookup against a None where it's expecting a Dict.
The fix is probably still correct, I think.

fnoop:
@mrpollo added a test that triggers the bug and fails in the current master, but passes after this PR is applied

mrpollo:
thanks for the clarification @fnoop I'm going to go ahead and merge, great work

Issue #568, opened 2016-02-18
Add copyright to all the example source files


hamishwillee:
As discussed in https://github.com/PX4/Firmware/pull/3790#issuecomment-185616039

brenton311:
Should the copyright notice be added to main source code along with the example source files?

hamishwillee:
Possibly. Not urgent though.

brenton311:
Added copyright notice to example files: https://github.com/dronekit/dronekit-python/pull/631

PR #631, opened 2016-05-26
Added copyright notice to example files.

brenton311:
Fixes issue at: https://github.com/dronekit/dronekit-python/issues/568


hamishwillee:
@mrpollo @peterbarker FYI. This is accurate for now and good to add. Arguably we should add additional copyright for dronecode if modifications are made?

@brenton311 Thanks Brenton, much appreciated.

PR #642, opened 2016-06-21
enable udpin-multi

mrpollo:
This enables udpin-multi mode, you can now setup dronekit to allow udpin from multiple sources, I have personally tested with up to 4, its not a hard limit its just what i was testing with, didn't really test with more but I'm sure we can support them as long as host has resources.

This feature is particularly special since it allows you to write a poor man's *`telem_forwarder`, this means we can now replicate some of its features in a cross-os environment w/o porting it.

*`telem_forwarder` is a service that lives inside 3DR Solo in between all of its responsibilities it allows you to write back to a `udpin` source from multiple clients and streams back mavlink to you from the port you started writing to.

@peterbarker I think you are going to want to review this since you have experience with `telem_forwarder`, this is specially important to us right now for development, let me know if you can't review.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,6 +20,8 @@ trim_trailing_whitespace = true
 [*.py]
 indent_style = space
 indent_size = 4
+trim_trailing_whitespace = true
+end_of_line = LF
 
 [*.mk]
 indent_style = tab
--- a/dronekit/mavlink.py
+++ b/dronekit/mavlink.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 import time
 import socket
+import errno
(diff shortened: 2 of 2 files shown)


peterbarker:
This looks fine, but wouldn't it be much better as patches to mavudp(mavfile)?  It's a modified copy of that function, and it seems the changes could be made in mavutil.py.

One thing I would be concerned about is aging stuff out of the addresses array.  That list will only grow in time.

mrpollo:
agreed, I think before 3.0 we should remove all of our mavutil enhancements and internal hacks and push upstream to pymavlink/mavutil

PR #646, opened 2016-07-14
Home position listener

georgehines:
Adds a listener to set the vehicle's home location when the autopilot broadcasts it (after it is set internally in the autopilot).

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1909,7 +1909,7 @@ class Vehicle(HasObservers):
         # Send MAVLink update.
         self.send_mavlink(self.message_factory.command_long_encode(
             0, 0,  # target system, target component
-            mavutil.mavlink.MAV_CMD_DO_SETH_OME,  # command
+            mavutil.mavlink.MAV_CMD_DO_SET_HOME,  # command
             0,  # confirmation
             2,  # param 1: 1 to use current position, 2 to use the entered values.
             0, 0, 0,  # params 2-4
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1200,6 +1200,11 @@ class Vehicle(HasObservers):
                 self._wploader.expected_count = msg.count
                 self._master.waypoint_request_send(0)
 
+
+        @self.on_message(['HOME_POSITION'])
+        def listener(self, name, msg):
+            self._home_location = LocationGlobal(msg.latitude/1.0e7, msg.longitude/1.0e7, msg.altitude/1000.0);
+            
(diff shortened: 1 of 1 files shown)


nick3dr:
This works as intended with Solo ShotManager. My only concern is that the _home_location also gets set on a MISSION_ITEM message, silently overwriting this.

Seems confusing. IMO the home location should be APM's home location and the waypoint_0 location should take another name. But this is up to someone who knows how this is used more than me and George.

@mrpollo can you please review this and merge with any necessary changes ASAP? Thanks!

hamishwillee:
@nick3dr @mrpollo My understanding is that the fact that the Waypoint 0 location is used as the home location is a historical bug, but unfortunately one that everyone now depends on. The actual value of both should be the same (though the home position message contains additional information about the approach vector). Using the HOME_POSITION message is a much better approach for dronekit because it doesn't require the download of (potentially large) missions in order to get the home position - and the consequently different behaviour for users when accessing this information. Of course it won't be supported in every autopilot build and we need to compensate for that. We also need to support requesting the value and setting the value, not just reading it. 

Therefore code will need to be added to:

- determine if there is support and conditionally disable the waypoint_0 based information
- add the approach vector information, setting values of attributes to None if this cannot be collected.
- send the getter message appropriately (I don't know when that should be, depends on whether autopilot sends it automatically)
- Change our setter method to set the value appropriately (assuming supported) rather than by writing to waypoint_0

nick3dr:
Thanks for the review @hamishwillee. Does any of that block this PR?

hamishwillee:
@nick3dr It shouldn't, provided the autopilot(s) don't stop populating waypoint_0 along with this change.

georgehines:
No functional changes to APM were made along with this change. APM sends
the HOME_POSITION message when it sets its internal home position in the
AHRS (at least in arducopter), so this listener merely picks up that
broadcast.

hamishwillee:
@georgehines Yes. My small concerns are that:

* someone will in future remove waypoint_0 storage of the value before proper checks are made on versioning ... on either ArduPilot or PX4 stacks. 
* The job will never be done through to completion

It's probably not worth delaying this being included.

mrpollo:
as @hamishwillee mentions theres no need to delay merge.

@peterbarker I'm going ahead and merging, but could you chime in on this thread with @hamishwillee and myself on what the proper way to handle backwards compat with APM/PX4 on issues like this is?, we might need a plan soon

hamishwillee:
@mrpollo @peterbarker The extra work required can be discussed in #649. It should be pretty easy.

PR #650, opened 2016-07-20
adds notification whenever home location is changed

hamishwillee:
Now we're updating the home_location from messages we can reasonably add attribute notification. Obviously this won't work before notification messages were added (we probably should add documentation to state that, rather than just saying "notification not supported".

This fixes minor bug in #646


hamishwillee:
@peterbarker @mrpollo Please review and push in.

mrpollo:
lgtm, :+1:

PR #653, opened 2016-08-22
fix a global_frame altitude caused bug in sitl test

tribbloid:
fix a bug where global_frame is used in case which global_relative_frame should be used instead, it causes the drone to take off using a very low sea-level altitude, so it won't leave the ground and its engine will be disarmed shortly after.


hamishwillee:
Thanks @tribbloid! Clearly a bug. Merged now.

PR #655, opened 2016-08-29
basic python 3 support

mrpollo:
we are using the `future` package to get proper `Queue` support for both python 2.7 and > 3.0.

Thanks @Rosuav for the patch on #604

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,4 +1,6 @@
 machine:
+  post:
+    - pyenv global 2.7.10 3.5.0
   environment:
     SITL_SPEEDUP: 10
     SITL_RATE: 200
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -40,7 +40,7 @@ import platform
 import re
 from dronekit.util import errprinter
 from pymavlink import mavutil, mavwp
-from Queue import Queue, Empty
+from queue import Queue, Empty
 from threading import Thread
 import types
 import threading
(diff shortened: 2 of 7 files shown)


hamishwillee:
It looks sane, but how is it tested?

mrpollo:
@hamishwillee I'm tweaking our CI configs so they run it on both Python versions, no need for more tests, the idea is that we should be able to run the same code on both versions until we drop one of them (hopefully 2.7 in the future)

hamishwillee:
OK. YOu'll have to re-merge. Perhaps @peterbarker should look at this too, but I'd be happy for you to self-merge :-)

mrpollo:
going ahead and merging, travis fails on a pymavlink python3 known problem, this has been raised on https://github.com/ArduPilot/pymavlink/issues/4 and will soon be fixed, until then we can't really claim full python 3 support on dronekit

circle.yml
```diff
@@ -1,4 +1,6 @@
 machine:
+  post:
+    - pyenv global 2.7.10 3.5.0
```
cclauss: Why not the current 2.7.12 and 3.5.2?

dronekit/__init__.py
```diff
@@ -40,7 +40,7 @@
 import re
 from dronekit.util import errprinter
 from pymavlink import mavutil, mavwp
-from Queue import Queue, Empty
+from queue import Queue, Empty
```
cclauss: Won't this break Python2 usage?

Normally, I would:

```python
try:                 # Python 3
    from queue import Queue, Empty
except ImportError:  # Python 2
    from Queue import Queue, Empty

```

PR #658, opened 2016-08-30
notify ci status to gitter

mrpollo:
both Travis and CircleCI


hamishwillee:
Any reason not to self merge? My only thought here was that I see a key marked as secure being pushed into a public repo :-)

mrpollo:
I'm self merging, just wanted to make sure i didn't brake CI, the secure key is already encrypted so we should be fine :)

PR #665, opened 2016-09-14
Gopro wifi

antonvh:
Some of this is also in another PR. What I did:
- listener on commands.next so I can get notified when a waypoint is reached
- camera class that connects to a GoPro3 over wifi (more cameras to come), and syncs time
- shutter() method that engages the shutter and logs a geotag.


dronekit/__init__.py
```diff
@@ -2748,6 +2794,65 @@ def __setitem__(self, index, value):
...
 
 
+class ServoOutputRaw(object):
+    """
+    The RAW servo readings from the OUTPUT of the autopilot
+    This contains the true raw values without any scaling to allow data capture and system debugging.
+
+    The message definition is here: https://pixhawk.ethz.ch/mavlink/#SERVO_OUTPUT_RAW
+
+
```
mrpollo: can you plz the extra new lines here?, and throughout if there are any other not needed newlines/spaces

dronekit/__init__.py
```diff
@@ -2748,6 +2794,65 @@ def __setitem__(self, index, value):
...
+    """
+
+    def __init__(self,
```
mrpollo: this is quite the signature, did you explore any other options to the servo attributes?, perhaps kwargs?

dronekit/__init__.py
```diff
@@ -2748,6 +2794,65 @@ def __setitem__(self, index, value):
...
+        self.servo16_raw = servo16_raw
+
+    def __str__(self):
+        """
+        String representation used to print the ServoOutputRaw object.
+        """
+        return "SERVO_OUTPUT_RAW: time_boot_us={},port={},servo1_raw={},servo2_raw={},servo3_raw={},servo4_raw={},servo5_raw={},servo6_raw={},servo7_raw={},servo8_raw={},servo9_raw={},servo10_raw={},servo11_raw={},servo12_raw={},servo13_raw={},servo14_raw={},servo15_raw={},servo16_raw={}".format(
+            self.time_boot_us, self.port, self.servo1_raw, self.servo2_raw, self.servo3_raw, self.servo4_raw,
+            self.servo5_raw, self.servo6_raw, self.servo7_raw, self.servo8_raw, self.servo9_raw, self.servo10_raw,
+            self.servo11_raw, self.servo12_raw, self.servo13_raw, self.servo14_raw, self.servo15_raw, self.servo16_raw)
```
mrpollo: We have been using the `__str__` decorator for debugging purposes on all of our top-level classes, do you think this is the most useful representation of data we can get out of this class?, would you typically expect all of the variables to be set and by reading make something out of it?, what if we are logging this, do you think this is the appropriate way to read it?

dronekit/__init__.py
mrpollo: thinking about it kwargs makes sense here, if you don't have all the servo messages in kwargs, you won't be able to set them on the internal attrs of the class, you would then need to create a helper perhaps to get the servo message you are looking for `.get(1)` or some sort of helper that makes sense (this is just an idea)

dronekit/cameras.py
```diff
@@ -0,0 +1,138 @@
...
+    def raise_timeout(self, *args):
+        raise Timeout.Timeout()
+
+
+class GeoTagger(object):
+    def __init__(self, vehicle=None):
+        self.vehicle = vehicle
+        FORMAT = '%(asctime)s, %(message)s'
+        DATEFORMAT = "%d-%m-%Y, %H:%M:%S"
+        logging.basicConfig(format=FORMAT, datefmt=DATEFORMAT, filename="camera_log.csv", filemode='w', level=logging.INFO)
```
mrpollo: i would expect the logging to be part of some other utility, or at least to be able to change the file name and formats used by it, what do you think?

dronekit/cameras.py
```diff
@@ -0,0 +1,138 @@
...
+        logging.info("picture number, waypoint, gimbal pitch, gimbal yaw, gimbal roll, att pitch, att yaw, att roll, latitude, longitude, altitude, camera message")
+
+    def log_vehicle_state(self, num_picture, cam_message):
+        if self.vehicle:
+            log_msg = ",".join(map(str,[num_picture, self.vehicle.commands.next,
```
mrpollo: I wonder if we should use the CSV lib for this, did you consider that when exploring this solution?

dronekit/cameras.py
```diff
@@ -0,0 +1,138 @@
...
+    try:
+        with Timeout(1):
+            data = urlopen(cmd).read()
+            return data
+    except Exception as e:
+            errprinter('>>> Exception in http command: ' + str(e))
+            return str(e)
+
+
+class GoProHero3(object):
```
mrpollo: I wonder if having a base `class Camera` object would be best here, you could inherit when creating a new camera object, that would allow us to maybe have a external modules, I would prefer that approach so we can standardize behind the same interfaces for cameras, this would allow for greater code re-use between apps.

dronekit/cameras.py
```diff
@@ -0,0 +1,138 @@
...
+        self.num_picture = 0
+        self.geotagger = None
+        if geotag:
+            self.geotagger = GeoTagger(vehicle=myvehicle)
+
+        # See https://github.com/KonradIT/goprowifihack/blob/master/WiFi-Commands.mkdn
+        # for a list of http commands to control the GoPro camera
+
+        #TODO: convert these into methods.
+        self.on = "http://10.5.5.9/bacpac/PW?t=" + self.wifipassword + "&p=%01"
```
mrpollo: can you create a url builder util for this class?, and perhaps add the IP as a variable too

dronekit/cameras.py
```diff
@@ -0,0 +1,138 @@
...
+        self.viewWide = "http://10.5.5.9/camera/FV?t=" + self.wifipassword + "&p=%00"
+        self.viewMedium = "http://10.5.5.9/camera/FV?t=" + self.wifipassword + "&p=%01"
+        self.viewNarrow = "http://10.5.5.9/camera/FV?t=" + self.wifipassword + "&p=%02"
+        self.res11mpWide = "http://10.5.5.9/camera/PR?t=" + self.wifipassword + "&p=%00"
+        self.res8mpMedium = "http://10.5.5.9/camera/PR?t=" + self.wifipassword + "&p=%01"
+        self.res5mpWide = "http://10.5.5.9/camera/PR?t=" + self.wifipassword + "&p=%02"
+        self.res5mpMedium = "http://10.5.5.9/camera/PR?t=" + self.wifipassword + "&p=%03"
+        self.noSound = "http://10.5.5.9/camera/BS?t=" + self.wifipassword + "&p=%00"
+        self.sound70 = "http://10.5.5.9/camera/BS?t=" + self.wifipassword + "&p=%01"
+        self.sound100 = "http://10.5.5.9/camera/BS?t=" + self.wifipassword + "&p=%02"
```
mrpollo: are you planning on adding all of these to the gopro interface? I wonder if we should have a generic `command` class which is handling all of them, and expose `shutter` as a top level user of `command('shutter')`?

dronekit/cameras.py
```diff
@@ -0,0 +1,138 @@
...
+                                        self.vehicle.location.global_frame.alt,
+                                        cam_message
+                                        ]))
+        else:
+            log_msg = str(num_picture)
+        logging.info(log_msg)
+        pass
+
+
+def send_http_cmd(cmd):
```
mrpollo: this should def live inside the gopro class

examples/mission_basic/mission_basic.py
```diff
@@ -1,216 +1,241 @@
-#!/usr/bin/env python
```
mrpollo: i think your gopro example needs its own example section, the mission_basic example isn't a good fit since not all the users are using or will use a gopro

mrpollo:
@peterbarker can you do some review of the APM interfaces and overall architecture plz?

@hamishwillee do you have any pointers for documentation?

kmk142789 (approved):

PR #672, opened 2016-10-07
pymavlink==2.2.2

cclauss:


cclauss:
Closed in favor of https://github.com/dronekit/dronekit-python/pull/673

peterbarker:
@cclauss @sanderux I've changed the requirements.txt away from a strict requirement of 2.0.6 to >= 2.0.6

I'm not sure why the requirement was so strict in the first place.  Some rather evil things had to be done because of that strict requirement when upstream removed 2.0.6....

@cclauss Why are you bumping the other package versions in this PR?

mrpollo:
just double checked, I installed today and i got 2.2.0 we should be fine

cclauss:
Thanks @mrpollo for your confirmation.  On the other two:
* `future` [aligns with upstream](https://github.com/ArduPilot/pymavlink/blob/0e949ae41c0db266c8a903a5bcc8d6d7e1bb5c20/requirements.txt#L2) leveraging the [good work that team](http://python-future.org/whatsnew.html#what-s-new-in-version-0-16-0-2016-10-27) is doing to iron out issues with running 2 and 3 together.
* `monotonic` brings us [timing support for AIX](https://github.com/atdt/monotonic/releases/tag/1.3) which is an [important platform](http://www-03.ibm.com/press/us/en/pressrelease/50688.wss) for my organization.

sanderux:
for px4 support, could we get pymavlink >= 2.2.2 ?

peterbarker:
Any reason not to >= for the monotonic support?

*double-take* on the AIX support.

peterbarker:
@sanderux I'm a bit concerned about pypi and versions getting installed.  So... why did @mrpollo get 2.2.0 instead of 2.2.2, for example?  Would not pay to break Ramon's laptop.

I think we should definitely move the requirement forward soon (PX4 being being a supported autopilot and all...), but I'd like to hear people naturally getting the requirement before we make that change.

If someone on MacOSX could `pip search pymavlink`, and someone on Windows `pip search pymavlink` and confirm that LATEST is 2.2.2, I think we could then move forward.

sanderux:
I think tridge might have released 2.2.0 as he mentioned to me he did. Perhaps 2.2.2 followed soon after.

Having the 2.2 requirement seems logical when people update dronekit. We could maybe ask tridge what the status is?

cclauss:
Looking at the [commit messages](https://github.com/ArduPilot/pymavlink/commits/master/setup.py), it appears that the bump of version numbers is the only difference between 2.2.0, 2.2.1, and 2.2.2.  And I quote "`darn pypi server isn't taking the uploads. 3rd try`".

I am fine with making the change to >= for monotonic support

peterbarker:
I've squished all these together and pushed them in - thanks!

@cclauss You have a fair point on those versions, but it still bothers me that some people aren't seeing recent versions of pymavlink.  If we start absolutely requiring those we may break things unnecessarily (new installs should work with PX4).

PR #673, opened 2016-10-07
 Python 3: define basestring, use bytes.decode()

cclauss:


mrpollo:
@cclauss I realize this is an old PR but we would benefit greatly from it, would you mind rebasing this and re submitting please?, I think tests should pass now

cclauss:
Closing in favor of #705

PR #679, opened 2016-11-03
adds a method to LocationLocal that returns distance from home, in me…

wsilva32:
…ters

I thought this would be a useful method to have on hand when trying to determine distance away from home. Includes a simple check that east, north, down are not None to avoid an exception.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -184,6 +184,13 @@ class LocationLocal(object):
     def __str__(self):
         return "LocationLocal:north=%s,east=%s,down=%s" % (self.north, self.east, self.down)
 
+    def distance(self):
+        """
+        Distance away from home, in meters.
+        """
+        if self.north is not None and self.east is not None and self.down is not None:
+            return math.sqrt(self.north**2 + self.east**2 + self.down**2)
+
 
 class GPSInfo(object):
     """
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -184,12 +184,16 @@ class LocationLocal(object):
     def __str__(self):
         return "LocationLocal:north=%s,east=%s,down=%s" % (self.north, self.east, self.down)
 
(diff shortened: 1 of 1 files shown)


dronekit/__init__.py
```diff
@@ -184,6 +184,13 @@ def __init__(self, north, east, down):
     def __str__(self):
         return "LocationLocal:north=%s,east=%s,down=%s" % (self.north, self.east, self.down)
 
+    def distance(self):
```
peterbarker: distance_home is probably a more suitable name; it would be in-line with https://github.com/ArduPilot/pymavlink/blob/master/mavextra.py#L501 for example.

dronekit/__init__.py
```diff
@@ -184,6 +184,13 @@ def __init__(self, north, east, down):
...
+        """
+        Distance away from home, in meters.
+        """
+        if self.north is not None and self.east is not None and self.down is not None:
```
peterbarker: Could we reformulate this so that if you don't have a component it isn't added instead?  That would allow someone who only has a 2D position to still work out where they are.

Kinda
```
sum = 0
if self.north is not None:
  sum += self.north**2
...
return math.sqrt(sum)
```

dronekit/__init__.py
wsilva32: Sure, I'm fine with that.

dronekit/__init__.py
wsilva32: makes sense, but perhaps only break it up between horizontal and vertical? I don't think we want to report a distance if it's only North or only East. I'll push a new commit

mrpollo:
failing un unrelated tests

hamishwillee:
@mrpollo It's been so long I can't remember if docs build automatically. In any case, in-source docs for this looked good. Can you remember how docs build worked?

PR #680, opened 2016-11-09
log all errprinter data using syslogd

wsilva32:
* expects a configuration file to exist at “/etc/shotmanager.conf”
* expects this file to contain a section called “dkpy”


wsilva32:
Whoops! This is solo specific and will be merged in 3DR's own fork.

PR #699, opened 2017-02-23
fix missing ** operator

gwen-le-mero:
in project pymavlink, all mavlink dialects define the send method (in their Mavlink class) as:
"def send(self, mavmsg, force_mavlink1=False):
    ...
"

In dronekit-pyhton project in file mavlink.py, there is a monkey patch on the mavlink send method to add extra processing (target fixing). fixed mavmsg is retransmitted back in original mavlink send function, but  unfortunately, all other parameters (kwargs) are transmitted back as dictionary params.

context: 
I found this bug while trying to activate mavlink2 between SITL and my dronekit script: Whatever i tried, sent mavlink messages where always in V1. With this fix i could achieve mavlink 2 communication

### DIFF
--- a/dronekit/mavlink.py
+++ b/dronekit/mavlink.py
@@ -140,7 +140,7 @@ class MAVConnection(object):
 
         def newsendfn(mavmsg, **kwargs):
             self.fix_targets(mavmsg)
-            return sendfn(mavmsg, kwargs)
+            return sendfn(mavmsg, **kwargs)
 
         self.master.mav.send = newsendfn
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@peterbarker Can you review this?

peterbarker:
@gwen-le-mero This looks good.  I've made the same patch myself at some stage.  I wonder where it got to...

Should be add *args in too - just in case upstream changes again?

gwen-le-mero:
followed your recommendations, and added the *args params (though i find it unlikely pymavlink will make breaking changes to their send function). 
I 'm a bit worried about failing tests, but it looks like its not my commit that made it break. (btw, i don't understand why 'basic python 3 support' was merged on master, if tests were failing for python 3.5 code.)

peterbarker:
Merged , thanks. And upstream already did change the function once, breaking dronekit-python :-)

PR #707, opened 2017-04-19
convert parameter set value to single precision floating point numbe…

amilcarlucas:
…r (the type used by low level mavlink messages) before sending/comparing it

This fixes bougus "timeout setting parameter x to y" messages when the passed double value and it's single representation differ


mrpollo:
great catch thanks!

PR #711, opened 2017-04-27
fix BaseException.message deprecated in 2.6 and now an error in 3.x

proximous:
BaseException.message was deprecated in 2.6 and now produces an Exception in 3.x.

Replace str(e.message) with str(e). Result should be the same in almost all cases.

See http://stackoverflow.com/a/19378402/3571110 for comments about potential differences.


peterbarker:
Merged, thanks!

PR #723, opened 2017-06-15
allow setting flight mode using a string value

larsks:
This modifies mode.setter so that we can write:

    vehicle.mode = 'LOITER'

Rather than:

    vehicle.mode = VehicleMode('LOITER')

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1561,7 +1561,26 @@ class Vehicle(HasObservers):
     @property
     def mode(self):
         """
-        This attribute is used to get and set the current flight mode (:py:class:`VehicleMode`).
+        This attribute is used to get and set the current flight mode. You
+        can specify the value as a :py:class:`VehicleMode`, like this:
+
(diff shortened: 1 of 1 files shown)


larsks:
@peterbarker does this seem like a reasonable idea?  Thanks!

peterbarker:
On Fri, 23 Jun 2017, Lars Kellogg-Stedman wrote:

> @peterbarker does this seem like a reasonable idea? Thanks!

Seems reasonable.

If you're happy with it then I can pull it in.  Any changes you'd like to 
make first?

(caveat: I thoroughly dislike the magically-caught-assignment stuff that 
you're modifying here; Hamish has a PR in place to replace it with 
something which sucks much less - attribute setters.  Since your approach 
here makes just as much sense with Hamish's work I'm still happy to see it 
go in :-) )

larsks:
Pull away!

peterbarker:
Tested and merged, thanks!

PR #730, opened 2017-06-29
fixing altitude float

HefnySco:
math.isnan(x) needs x to be float.
the code uses alt instead of altitude


peterbarker:
Merged, thanks!

PR #737, opened 2017-07-23
Ignore heartbeats from groundstations

sanderux:
When connected over a network using broadcast dronekit gets confused when receiving heartbeats from a groundstation.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1189,19 +1189,21 @@ class Vehicle(HasObservers):
 
         @self.on_message('HEARTBEAT')
         def listener(self, name, m):
-            self._armed = (m.base_mode & mavutil.mavlink.MAV_MODE_FLAG_SAFETY_ARMED) != 0
-            self.notify_attribute_listeners('armed', self.armed, cache=True)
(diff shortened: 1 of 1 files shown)


hamishwillee:
@peterbarker Is this going to conflict with the "proper" support for multiple mavlink sources you put in?

peterbarker:
@hamishwillee Not really, but it could/should be removed if we ever actually get my stuff in.  So I think we should put this in - I can't see how it will hurt anybody, and it would seem to help at least somebody :-)

@sanderux I'm not convinced this is a complete solution; a similar problem came up to do with MAVProxy and pymavlink; this is the same sort of check: https://github.com/ArduPilot/pymavlink/compare/master...peterbarker:heartbeat-changes?expand=1  Note that gimbals are also known to have heartbeats...

@sanderux Could you use a return-early rather than reindenting everything, please?  Since there's been two recent issues with this multiple-mavlink-sources-confusing-everything-and-everybody, I think I really need to come back to the pymavlink mavsource stuff - so hopefully these patches won't be required for too long!

peterbarker:
Squashed commits together and merged them.

Thanks!

PR #740, opened 2017-07-24
Add access to attitude rates 

cheintz:
This seemed to be a good place for them, as they come in the Attitude MAVlink packet.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -89,9 +89,9 @@ class Attitude(object):
         self.pitch = pitch
         self.yaw = yaw
         self.roll = roll
-	self.pitchspeed = pitchspeed
-	self.yawspeed = yawspeed
-	self.rollspeed = rollspeed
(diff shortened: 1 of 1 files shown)


peterbarker:
I think we should have separate objects and accessors for the rotation rates.

We don't conflate position and its first derivation velocity, so I can't see why we'd want to do it for attitude.

So basically an AttitudeRate object.

Does that make sense?

Ah, the argument "that's how MAVLink does it" isn't a great one, either.  MAVLink (and its implementations) gets lots of stuff wrong, and dronekit-python is supposed to kinda fix these things :-)

hamishwillee:
FWIW Yes to an `AttitudeRate` object. Would be good if during implementation this could be added to the [vehicle_state.py](https://github.com/dronekit/dronekit-python/blob/master/examples/vehicle_state/vehicle_state.py) as the absolute bare minimum test.

amilcarlucas:
@cheintz any progress on this ?

PR #743, opened 2017-08-01
isflying added

HefnySco:
implement test criteria to know if a vehicle is flying or not.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.linting.pylintEnabled": false
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.linting.pylintEnabled": false
+}
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1983,7 +1983,7 @@ class Vehicle(HasObservers):
         """
         if alt is not None:
             altitude = float(alt)
-            if math.isnan(alt) or math.isinf(alt):
+            if math.isnan(altitude) or math.isinf(altitude):
                 raise ValueError("Altitude was NaN or Infinity. Please provide a real number")
             self._master.mav.command_long_send(0, 0, mavutil.mavlink.MAV_CMD_NAV_TAKEOFF,
                                                   0, 0, 0, 0, 0, 0, 0, altitude)
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1067,6 +1067,7 @@ class Vehicle(HasObservers):
         self._heading = None
         self._airspeed = None
         self._groundspeed = None
+        self._isflying = False
 
         @self.on_message('VFR_HUD')
         def listener(self, name, m):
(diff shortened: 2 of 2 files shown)


peterbarker:
Terminology problem here.

None of the checks you've added here have anything to do with flying.  A vehicle could pass all of those checks and be an AntennaTracker :-)

Determining whether a vehicle is in the air under its own steam is actually really quite hard.  You really need to ask it what the *probability* of it being actively flying itself currently is.  ArduPilot currently has such a concept for both Plane, and something related in Copter.

What were you hoping this particular check would represent about the vehicle?

HefnySco:
You are correct :+1:  .... It is a hard problem.
It is up to developer to use this feature based on vehicle types, it is similar like switching to Surface mode and your vehicle is a plane not a sub.  However when vehicle type is plane or drone it works fine. I already use in in DronekitAPI in the Android Version.

**isFlying** is a variable already exists in Dronekit for Android [here](https://github.com/dronekit/dronekit-android/search?utf8=%E2%9C%93&q=isflying&type=)

hamishwillee:
>  ArduPilot currently has such a concept for both Plane, and something related in Copter.

@peterbarker Is there access to the test via a mavlink message? If so, that could be the check, and the rest could be some sort of fallback for other platforms. 

> It is up to developer to use this feature based on vehicle types, it is similar like switching to Surface mode and your vehicle is a plane not a sub. However when vehicle type is plane or drone it works fine. I already use in in DronekitAPI in the Android Version.

Dronekit isn't supposed to work like that. If you want to have ambiguous magic behaviour then ideally you'd define a superclass for that vehicle. Minimally you'd want to check the vehicle and make sure an error of some kind is returned for vehicles that don't fly.

HefnySco:
@hamishwillee Do you suggest limiting it to plane and copter as follows:

      @property
      def isflying (self):
        if((self._vehicle_type != mavutil.mavlink.MAV_TYPE_QUADROTOR) and
            (self._vehicle_type != mavutil.mavlink.MAV_TYPE_FIXED_WING)):

            return None

        isFlying = ((self._system_status == Vehicle.MAV_STATE_ACTIVE) or 
            ((self._isflying == True) and (self._system_status == Vehicle.MAV_STATE_CRITICAL or Vehicle._system_status == self.MAV_STATE_EMERGENCY)))

        self._isflying = isFlying
                
        
            


        return self._isflying

hamishwillee:
@HefnySco I'd really defer to @peterbarker , since he is far more involved with me.  This is the sort of thing I was thinking. A good alternative might be to have a more generic method - e.g.  "isLaunched" or "isMoving" rather than isFlying, and have test conditions that also make sense for ground vehicles and subs.

I don't know the right tests to have in place to show something is flying/moving though.

pietrodn:
Alternative idea... it may be not exactly what OP needs, but I had success with it on my drones, and it could give valuable information to users.

In DroneKit, we could monitor the [`EXTENDED_SYS_STATE`](https://mavlink.io/en/messages/common.html#EXTENDED_SYS_STATE) MAVLink message to determine if the `landed_state` attribute is equal to [`MAV_LANDED_STATE_ON_GROUND`](https://mavlink.io/en/messages/common.html#MAV_LANDED_STATE_ON_GROUND).

In PX4, `landed_state` is equal to `MAV_LANDED_STATE_ON_GROUND` if and only if the [land detector Finite State Machine](https://docs.px4.io/en/advanced_config/land_detector.html) is in the "Landed" state.

Clearly, it's important to document that such an attribute gives only access to the land detector state, and it may not represent the actual drone status (e.g. if the engines are "twitching", the land detector may fail to detect ground contact).

PR #749, opened 2017-09-04
Avoid runtime NameError by defining `time_max` in sitl/test_110.py

cclauss:
Without this change, a NameError is raised on line 50.

https://travis-ci.org/dronekit/dronekit-python/jobs/271892375#L606-L608

### DIFF
--- a/dronekit/test/sitl/test_110.py
+++ b/dronekit/test/sitl/test_110.py
@@ -43,7 +43,8 @@ def test_110(connpath):
     vehicle.armed = True
 
     # Wait for ACK.
-    wait_for(lambda : armed_callback.called, 10)
+    time_max = 10
+    wait_for(lambda : armed_callback.called, time_max)
 
     # Ensure the callback was called.
     assert armed_callback.called > 0, "Callback should have been called within %d seconds" % (time_max,)


dronekit/test/sitl/test_110.py
```diff
@@ -43,7 +43,8 @@ def armed_callback(vehicle, attribute, value):
     vehicle.armed = True
 
     # Wait for ACK.
-    wait_for(lambda : armed_callback.called, 10)
+    time_max = 10
+    wait_for(lambda : armed_callback.called, time_max)
```
peterbarker: This parameter isn't actually a time; it's an arbitrary value which `armed_callback.called` must reach.  It looks like this test probably used to have a time requirement but it was removed.  Suggest you could create a "required_call_count" variable instead and change the assert message to say something about should have been called %u times...

cclauss: Thanks. Peter, `wait_for()` is defined here: https://github.com/dronekit/dronekit-python/blob/master/dronekit/test/__init__.py#L31-L34  It uses the same variable name for its 2nd parameter, `time_max` which is counted off of `time.time()`.

This means that as advertised in the assert text this test will wait for a max of 10 seconds to see if armed_callback.called goes positive.  Perhaps `max_seconds`, or `seconds_to_wait` would be a more intuitive variable name?  If we make a variable name change, do we just do it here or do we also make a similar change in the definition of `wait_for()`?

peterbarker: My bad.

PR #752, opened 2017-09-14
Modernize Python 2 code to get ready for Python 3

cclauss:
This PR replaces #750 and uses http://python-future.org code to do a more complete job than 750.

Make the minimal, safe changes required to convert the code to be syntax compatible with both Python 2 and Python 3. There would definitely be [__more work required__](https://travis-ci.org/dronekit/dronekit-python/jobs/275427857#L1608) to complete the port of the code to Python 3 but this is a minimal, safe first step.

1. Run: `futurize --stage1 -w **/*.py`

See Stage 1: "safe" fixes http://python-future.org/automatic_conversion.html#stage-1-safe-fixes

$ `pip install future`
$ `git checkout -b modernize-python2-code`
$ `futurize --stage1 -w **/*.py`  # done in zsh for ** globbing
$ `git commit --all -m "Modernize Python 2 code to get ready for Python 3"`
$ `git push --set-upstream origin modernize-python2-code`

2. Remove the need for xrange() in `examples/flight_replay/flight_replay.py`
3. Do better testing on Python 3 in allow_failures mode in `.travis.yml`

PR #758, opened 2017-10-12
Working Py3 code that still supports Py2

cclauss:
__futurize --stage2 **/*.py__
* http://python-future.org/futurize_cheatsheet.html#step-2-working-py3-code-that-still-supports-py2

The first stage of the futurize process was done in #752 and this is the second (and final) stage of that process.


cclauss:
Travis refuses to test this PR.

PR #762, opened 2017-11-01
send_capabilities_request was misspelled

larsks:
this corrects the spelling of "capabilities" and adds an alias to the
old name to avoid breaking existing code.


peterbarker:
Merged, thanks!

PR #765, opened 2017-11-03
raise ValueError if we are passed a bad value

larsks:
Given a bad value, we were sometimes raising a bare Exception,
sometimes raising a ValueError, and sometimes raising an APIError.
This commit attempts to make things consistent by raising ValueError
when that's what we mean.

There is also a drive-by correction to the spelling of "Expecting".

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -730,7 +730,7 @@ class ChannelsOverride(dict):
 
     def __setitem__(self, key, value):
         if not (int(key) > 0 and int(key) <= self._count):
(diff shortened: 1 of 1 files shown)


larsks:
...and why is it a bad idea to raise a bare `Exception`? Because the only way to catch that is with:

    except Exception

...which will catch *everything*, including your syntax errors, `AttributeErrors`, and a variety of other common problems that get masked by this sort of thing.

peterbarker:
Merged, thanks!

PR #766, opened 2017-11-03
add convenience methods for common synchronous operations

larsks:
this adds wait_for_armable, arm, disarm, wait_for_mode, and wait_for_alt
methods that wrap common operations with wait-for-complete behavior
with an optional timeout.


peterbarker:
About time - thanks!

PR #768, opened 2017-11-09
Add still-waiting interval and callback to wait_ready

peterbarker:


williamorpheus:
interesting, this will be really useful... keep up the good work

amilcarlucas:
looks good, but I can only test in January. Any brave testes out there.

PR #784, opened 2018-01-12
Python 3 fixes to make tests succeed

pietrodn:
I made some fixes to the codebase so that the SITL tests succeed when running on Python 3 (now they fail).

I also ran the tests on Python 2 and they didn't break.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1290,7 +1290,7 @@ class Vehicle(HasObservers):
                 c = 0
                 for i, v in enumerate(self._params_set):
                     if v == None:
-                        self._master.mav.param_request_read_send(0, 0, '', i)
+                        self._master.mav.param_request_read_send(0, 0, b'', i)
                         c += 1
                         if c > 50:
                             break
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -197,7 +197,7 @@ class LocationLocal(object):
 
         if self.north is not None and self.east is not None:
             if self.down is not None:
-                return math.sqrt(self.north**2 + self.east**2 + self.down**2) 
+                return math.sqrt(self.north**2 + self.east**2 + self.down**2)
             else:
                 return math.sqrt(self.north**2 + self.east**2)
 
@@ -342,7 +342,7 @@ class Version(object):
         if self.release is None:
             return None
         types = [ "dev", "alpha", "beta", "rc" ]
-        return types[self.release/64]
+        return types[self.release//64]
 
     def __str__(self):
         prefix=""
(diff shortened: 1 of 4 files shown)


pietrodn:
To make the tests run on Python 3, pymavlink also needs to be patched. I just [submitted](https://github.com/ArduPilot/pymavlink/pull/138) the patch on the pymavlink repo.

dronekit/__init__.py
```diff
@@ -342,7 +342,7 @@ def release_type(self):
         if self.release is None:
             return None
         types = [ "dev", "alpha", "beta", "rc" ]
-        return types[self.release/64]
+        return types[self.release//64]
```
peterbarker: This really should be a shift.

dronekit/__init__.py
```diff
@@ -2658,8 +2658,7 @@ def thr_min_callback(self, attr_name, value):
         :param args: The callback to invoke when a change in the parameter is detected.
 
         """
-        attr_name = attr_name.upper()
-        return super(Parameters, self).add_attribute_listener(attr_name, *args, **kwargs)
+        return super(Parameters, self).add_attribute_listener(attr_name.upper(), *args, **kwargs)
```
peterbarker: Mmmm.... this doesn't look like python3 support to me.

dronekit/test/sitl/test_mavlink.py
```diff
@@ -24,5 +24,6 @@ def callback(*args):
     i = 20
     while not result['success'] and i > 0:
         time.sleep(1)
+        i -= 1
```
peterbarker: Yowsers.

Not a python3 thing, 'though :-)

dronekit/__init__.py
```diff
@@ -1316,7 +1316,7 @@ def listener(self, name, msg):
                         self._params_last = monotonic.monotonic()
                         self._params_duration = start_duration
                     self._params_set[msg.param_index] = msg
-                self._params_map[msg.param_id] = msg.param_value
+                self._params_map[msg.param_id.rstrip(b'\x00').decode('utf8')] = msg.param_value
```
peterbarker: I wonder if we should be stripping these at the pymavlink level.

dronekit/__init__.py
pietrodn: Uhm, I think I changed this to print the parameter name while debugging and I forgot to change it back. Should be reverted (will do it).

dronekit/__init__.py
pietrodn: Probably yes. I will look into it.

pietrodn:
Superseded by #788 and #789. Closing.

PR #788, opened 2018-01-22
Fixes and improvements for two SITL tests

pietrodn:
Sometimes the SITL tests in `test_mavlink.py` fail because the vehicles are not ready.
We can avoid this by adding `wait_ready=True` to the `Vehicle.connect()` call and obtain more reliable result.

Also, a timeout condition would never activate because a counter was not decreased when it should have to.

In `test_parameters.py`, I added parentheses to make the assignment more legible.

This pull request contains the non-Python3-compatibility fixes from #784.


peterbarker:
Merged, thanks!

PR #798, opened 2018-02-12
Fix param read loop for python 3

tpwrules:
This commit changes the 'blank parameter ID' value from empty string to empty bytes, so that it can be packed for transmission to the copter in Python 3.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1166,6 +1166,9 @@ class Vehicle(HasObservers):
         @self.on_message(['WAYPOINT_CURRENT', 'MISSION_CURRENT'])
         def listener(self, name, m):
             self._current_waypoint = m.seq
+            if m.seq != self._current_waypoint:
+                 self._current_waypoint = m.seq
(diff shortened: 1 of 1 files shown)


pietrodn:
I think this should be fixed in pymavlink rather than in DroneKit: the correct type for a parameter name should be a `str`, not `bytes`.
I already submitted a PR in pymavlink which would fix this error: https://github.com/ArduPilot/pymavlink/pull/141

amilcarlucas:
There have been some recent changes on pymavlink regarding this. Can you retest?

pietrodn:
The first commit is OK as the `param_request_read_send` [accepts a byte string](https://github.com/ArduPilot/pymavlink/blob/670cc9a786666e0e7931f93c36242e45d0554ca2/mavutil.py#L412) as the third argument.
The second commit is unrelated and should be stripped from the PR.

pietrodn:
This can be closed in favor of #865, as this contains an unrelated change and needs a rebase.

amilcarlucas:
@peterbarker ardupilot/pymavlink#865 has been merged, so this one can be closed now

PR #807, opened 2018-03-25
circle.yml: use latest available Python versions

pietrodn:
Currently the CircleCI build is [failing](https://circleci.com/gh/dronekit/dronekit-python/1341) because Python 2.7.13 is not installed on the system image. Here is a list of available Python versions: https://circleci.com/docs/1.0/build-image-trusty/#python

This PR should fix that problem.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   post:
-    - pyenv global 2.7.13 3.6.0
+    - pyenv global 2.7.12 3.6.2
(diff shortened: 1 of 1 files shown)


peterbarker:
Merged, thanks. 

Can't get much worse, after all :-)

pietrodn:
Still doesn’t work... which is the build environment? Is it Ubuntu 14.04?

PR #809, opened 2018-03-26
CircleCI: fix the build (again...)

pietrodn:
- Install future beforehand as workaround to pymavlink bug: https://github.com/ArduPilot/pymavlink/issues/25
- Install dependencies before dronekit, not after

I tested this on my CircleCI so it should work. :-)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -7,8 +7,9 @@ machine:
 
 dependencies:
   override:
-    - 'pip install . -UI'
+    - 'pip install -UI future'
     - 'pip install -r requirements.txt -UI'
+    - 'pip install . -UI'
 
 test:
   override:


peterbarker:
On Mon, 26 Mar 2018, Pietro De Nicolao wrote:

> Now it fails on my CircleCI because I'm using Ubuntu 14.04, which DroneKit's
> CircleCI is using 12.02 (I think). Should be fine though.

Are you suggesting that the current failure,

----
pyenv global 2.7.10 3.5.0

pyenv: version `3.5.0' not installed

pyenv global 2.7.10 3.5.0 returned exit code 1

Action failed: pyenv global 2.7.10 3.5.0
----

will be fixed once this is merged?

pietrodn:
Yup. The current master has the very same Python versions and it finds them.

peterbarker:
Well then - let's give that a shot!

PR #811, opened 2018-03-31
Autopilot reboot + test

pietrodn:
This PR implements a simple to use `Vehicle.reboot()` command to reboot the autopilot via software by sending a `MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN`.

I tested it on a Pixfalcon running PX4 and it works as expected.
The SITL doesn't support a real "reboot", but it still ACKs the command to the client. I added a test for that.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -2360,6 +2360,21 @@ class Vehicle(HasObservers):
 
         return True
 
+    def reboot(self):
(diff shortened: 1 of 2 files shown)


peterbarker:
Added an example (so I could test it!  Works on ArduCopter).

Merged, thanks!

pietrodn:
Great! Thank you :)

PR #814, opened 2018-04-11
disable docs deploy from circleci

mrpollo:
we are migrating service providers and need to disable deployments of
docs


hamishwillee:
Where are docs going to?

PR #828, opened 2018-05-15
Home location set command corrected, fixes #827

### LINKED ISSUE #827: Home location setter forms invalid command
According to the message definitions in [MAVLINK](http://mavlink.org/messages/common), message 179 should use a value of 0 in param1 to set to entered values.

The home location setter uses a value of 2, which is ignored by Ardupilot.

michael-hart:
param1 should have value 0 rather than 2

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1969,7 +1969,7 @@ class Vehicle(HasObservers):
             0, 0,  # target system, target component
             mavutil.mavlink.MAV_CMD_DO_SET_HOME,  # command
             0,  # confirmation
-            2,  # param 1: 1 to use current position, 2 to use the entered values.
+            0,  # param 1: 1 to use current position, 0 to use the entered values.
             0, 0, 0,  # params 2-4
             pos.lat, pos.lon, pos.alt))
 


peterbarker:
On Tue, 15 May 2018, michael-hart wrote:

> param1 should have value 0 rather than 2

Well, that's odd.

Can you check that this wasn't important behaviour for PX4 Firmware, 
please?

michael-hart:
I found this behaving incorrectly when trying to set the home location on my PixHawk Cube. When it didn't work, I investigated in simulator, attaching GDB and stepping through as it tried to set the home location - it was ignored because param1 was not either 0 or 1. Once I made the change in my branch and installed that using pip, the home location set correctly

pietrodn:
The [MAVLink specification](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_HOME) says that to use the location specified in the message as the new home location you need to set param1 = 0.

PX4 uses the current location if param1 > 0.5, and the location specified in the message if param1 <= 0.5 ([here's the code](https://github.com/PX4/Firmware/blob/master/src/modules/commander/Commander.cpp#L798)), so it is compliant.

hamishwillee:
Thank you @michael-hart @pietrodn . You are absolutely correct. Very odd that this was incorrect. 

@peterbarker Is there any test for this (or should there be one?_

PR #840, opened 2018-08-02
__init__.py: rename variable "await"

pietrodn:
As of Python 3.7, "await" is a reserved word ([PEP 492](https://www.python.org/dev/peps/pep-0492/#await-expression)).


peterbarker:
Merged, thanks!

PR #845, opened 2018-08-22
Fixed the status text not showing up on python3

kaangoksal:
I'm using dronekit with python3 and I was getting lots of:

>>> Exception in message handler for STATUSTEXT
>>> cannot use a string pattern on a bytes-like object
>>> Exception in message handler for STATUSTEXT
>>> cannot use a string pattern on a bytes-like object
>>> Exception in message handler for STATUSTEXT
>>> cannot use a string pattern on a bytes-like object
>>> Exception in message handler for STATUSTEXT
>>> cannot use a string pattern on a bytes-like object

instead of:

>>> APM:Copter V3.3 (d6053245)
>>> Frame: QUAD
...

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -3012,7 +3012,7 @@ def connect(ip,
 
         @vehicle.on_message('STATUSTEXT')
         def listener(self, name, m):
-            status_printer(re.sub(r'(^|\n)', '>>> ', m.text.decode("utf-8").rstrip()))
+            status_printer(re.sub(r'(^|\n)', '>>> ', m.text.rstrip()))
 
     if _initialize:
         vehicle.initialize(rate=rate, heartbeat_timeout=heartbeat_timeout)


dronekit/__init__.py
```diff
@@ -3012,7 +3012,7 @@ def connect(ip,
 
         @vehicle.on_message('STATUSTEXT')
         def listener(self, name, m):
-            status_printer(re.sub(r'(^|\n)', '>>> ', m.text.rstrip()))
+            status_printer(re.sub(r'(^|\n)', '>>> ', m.text.decode("utf-8").rstrip()))
```
peterbarker: There's no guarantee that m.text is valid unicode data.  In fact, I'm investigating a case ATM where it is not (but should be)!

I'm concerned that after this patch we might end up with failed-to-decode exceptions of some kind here.

I think we'll have to accept that we can't use a regex here and find some other mechanism to do the sub - perhaps even just assume all statustext lines are single-line-only (which you kind of have to if you don't know what the text encoding is)!

amilcarlucas:
There have been some pymavlink recent changes in this area. Can you retest, it might be fixed now.

kaangoksal:
Will look into it in the weekend, thanks for the update

kaangoksal:
Okay, yeah my problem is solved it works perfectly.

amilcarlucas:
Can you please rebase this on top of master?

amilcarlucas:
A rebase would be nice anyways. There seams to be some good things in this PR.

pietrodn:
I can't even see the original fix in the diff. It looks like some commits which were already merged leaked in the PR.

PR #860, opened 2018-10-20
Properly close Vehicle instances after SITL tests

pietrodn:
If Vehicle instances are left unclosed, the latest pymavlink (2.2.20) will spam a lot of "EOF on TCP socket" warning messages:
https://github.com/ArduPilot/pymavlink/issues/230

This commit fixes the issue and allows DroneKit to be used with the latest pymavlink.
In turn, this will enable full Python 3 compatibility for DroneKit. :-)


peterbarker:
Merged, thanks!

Issue #863, opened 2018-10-27
High latency in MAVLink message receiving


pietrodn:
I wrote a simple script to measure the MAVLink message latency (round-trip time) by exploiting the `TIMESYNC` protocol as implemented in PX4.
The protocol works like this: in the client application you send a `TIMESYNC` message to the autopilot, with:
* `tc1=0`;
* `ts1` set to the locally generated timestamp.

The firmware responds ([source code](https://github.com/PX4/Firmware/blob/6f204c491335af88dcf1c13484e3d98f82868296/src/modules/mavlink/mavlink_timesync.cpp#L67)) with another `TIMESYNC`, having:
* the same `ts1` field (the timestamp of the original message);
* the autopilot's timestamp in the `tc1` field.

By observing the difference between the timestamp of arrival of the second message and its `ts1` field, you can effectively measure the round-trip time of the message and estimate the link latency (e.g. by assuming latency = RTT / 2).
You can also estimate the offset between the client and the autopilot's clocks using the `tc1` field, but I'm not interested in this right now.

I did the following experiment: send a `TIMESYNC` to the autopilot every 10 ms for 100 times, and measure the round-trip times from the answers.
I'm using the PX4 SITL simulator, so I'm communicating with the autopilot through UDP on the same host (ideal conditions, no packet loss).

I noticed a strange pattern in the measured latencies (**orange line** in the graph).

![timesync latencies rtt _ 2](https://user-images.githubusercontent.com/5058646/47603124-c1638e00-d9e7-11e8-80ab-71ae1e8cffde.png)

In the graph we can observe groups of 4-5 messages with strictly decreasing latencies, starting from 25 ms  down to about 5 ms or less.

I also observed the autopilot-generated timestamps (`tc1` field) and noticed that they were very regular (there were about 10 ms between them, always).
So, this must be a problem on the receiving side of the client.

The cause of the problem is [this line of code](https://github.com/dronekit/dronekit-python/blob/5d2614f79092cc3ea03062a084579eb16577cd22/dronekit/mavlink.py#L207): `time.sleep(0.05)` in the `mavlink_thread_in` function.

That `sleep()` makes DroneKit wait 50 ms if no messages are found in the buffer; it matches the 25 ms delay shown in the graph (I'm plotting RTT / 2 as a latency estimate).

By removing the sleep, I get dramatically lower latencies (less than 1 ms on average): they are shown on the **blue line** in the above graph.

Why is that sleep there? Is it possible to remove it?
[...]

PR #864, opened 2018-10-28
mavlink.py: efficient I/O waiting with select()

pietrodn:
pymavlink uses non-blocking [sockets](https://github.com/ArduPilot/pymavlink/blob/master/mavutil.py#L926) and [serial ports](https://github.com/ArduPilot/pymavlink/blob/master/mavutil.py#L824): polling for data in a while loop with a `sleep(0.05)` is inefficient and leads to increased latency for incoming MAVLink messages (issue #863).

But, if the `sleep` call is blindly removed, then the CPU usage rises to 100% (busy-looping)!

The [`select()`](https://docs.python.org/3/library/select.html#select.select) system call, [exposed in the pymavlink API](https://github.com/ArduPilot/pymavlink/blob/master/mavutil.py#L209), puts the `mavlink_thread_in` thread to sleep for just the right amount of time and wakes it up when fresh data is ready to be processed.

If `select()` is used, the returned message may never be `None`, so the loop listeners would never get called. This is also fixed by ensuring that the loop listeners are called at least every 0.5 s.


peterbarker:
I like the look of this, but we shouldn't massively increase the jitter on calling those functions.

pietrodn:
Well, currently the loop listeners are called at an arbitrary, unknown interval: if the input buffer is always full, they may as well never get called after the first loop.

What would be a "correct" frequency for those functions?

pietrodn:
Ok, I figured out that after each select (or sleep) we need to consume all messages until `recv_msg()` returns `None`. And then execute the loop listeners.

peterbarker:
On Sun, 28 Oct 2018, Pietro De Nicolao wrote:

> Well, currently the loop listeners are called at an arbitrary, unknown
> interval: if the input buffer is always full, they may as well never get
> called after the first loop.

This is true - but I'd *guess* an unlikely scenario.

> What would be a "correct" frequency for those functions?

I'm guessing 20Hz would be reasonable - so a select delay of .05

peterbarker:
Could you squash this and fix the commit message, please?

This is a good change.

pietrodn:
Thanks! Just squashed the commits.

PR #866, opened 2018-10-30
Performance improvements (comparisons with None, True, False)

pietrodn:
While profiling my DroneKit-based application with yappi, I noticed that the [`MAVLink_message.__eq__()`](https://github.com/ArduPilot/pymavlink/blob/eaa2ab77e9490f0427d86a2401ab61c99b6bca2b/generator/mavgen_python.py#L165) method was called more than 1.9 million times over a few minutes.
It was the function called with the greatest frequency in the whole application, and consumed a lot of CPU time.
QCacheGrind: functions by number of calls

I inspected the call stack using QCacheGrind and found out that all of these calls were comparisons of a MAVLink message object with `None`. All of them can be avoided by using the idiomatic constructs `is None` / `is not None`.

Some of these calls (~13k) were from pymavlink: these [are already gone](https://github.com/ArduPilot/pymavlink/pull/235).

Callers subdivision

All the other calls are coming from DroneKit, specifically from [this line of code](https://github.com/dronekit/dronekit-python/blob/4e392d603818ff29eadf6518122c6418d6c67f95/dronekit/__init__.py#L1285) inside a loop listener (a function executed once every few incoming MAVLink message):
```python
        # list containing 1 item for each parameter of the autopilot:
        # each can be None (param not loaded) or a PARAM_VALUE message object
        self._params_set = []
        self._params_loaded = False

        @handler.forward_loop
        def listener(_):
            # HERE: "not in" calls MAVLink_message.__eq__()
            # for each non-None element in self._params_set
            if None not in self._params_set and not self._params_loaded:
                self._params_loaded = True
                self.notify_attribute_listeners('parameters', self.parameters)
```

That code is expensive. The current version of PX4 has more than 500 parameters: even after all the parameters are loaded, on every message loop, DroneKit calls `MAVLink_message.__eq__()` 500+ times!

I improved the code:
```python
            # Exploit short-circuiting of the and operator (check the boolean first)
            # so that, if self._params_loaded == True, the array does not get checked;
            # replace the expensive `__eq__()` call with a fast, idiomatic
            # `is not None` check
            if not self._params_loaded and all(x is not None for x in self._params_set):
```

I went ahead and fixed all the comparisons with `None`, `True`,  `False`, and replaced a double lookup in a dictionary with a `.get()` call.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -304,7 +304,7 @@ class Version(object):
         self.autopilot_type = autopilot_type
         self.vehicle_type = vehicle_type
         self.raw_version = raw_version
-        if(raw_version == None):
+        if raw_version is None:
             self.major = None
             self.minor = None
             self.patch = None
@@ -640,7 +640,7 @@ class HasObservers(object):
         """
         # Cached values are not re-sent if they are unchanged.
         if cache:
-            if attr_name in self._attribute_cache and self._attribute_cache[attr_name] == value:
+            if self._attribute_cache.get(attr_name) == value:
                 return
             self._attribute_cache[attr_name] = value
 
@@ -885,7 +885,7 @@ class Locations(HasObservers):
             vehicle.notify_attribute_listeners('location.global_relative_frame',
                                                vehicle.location.global_relative_frame)
 
-            if self._alt != None or m.alt != 0:
+            if self._alt is not None or m.alt != 0:
                 # Require first alt value to be non-0
                 # TODO is this the proper check to do?
                 self._alt = m.alt / 1000.0
@@ -1197,7 +1197,7 @@ class Vehicle(HasObservers):
             self.notify_attribute_listeners('armed', self.armed, cache=True)
             self._autopilot_type = m.autopilot
             self._vehicle_type = m.type
-            if self._is_mode_available(m.custom_mode, m.base_mode) == False:
+            if self._is_mode_available(m.custom_mode, m.base_mode) is False:
                 raise APIException("mode (%s, %s) not available on mavlink definition" % (m.custom_mode, m.base_mode))
             if self._autopilot_type == mavutil.mavlink.MAV_AUTOPILOT_PX4:
                 self._flightmode = mavutil.interpret_px4_mode(m.base_mode, m.custom_mode)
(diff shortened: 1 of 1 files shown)


hamishwillee:
@pietrodn Nice one. Just for my interest, what was the profiled performance improvement?

@peterbarker Best you review this one.

dronekit/__init__.py
```diff
@@ -1282,7 +1282,7 @@ def listener(_):
             if not self._params_start:
                 return
 
-            if None not in self._params_set and not self._params_loaded:
+            if not self._params_loaded and all(x is not None for x in self._params_set):
```
peterbarker: Is this correct?  Is it actually faster?

pietrodn:
I re-profiled my program with and without this patch using the PX4 SITL.

Without the patch, the list of the methods occupying the most CPU time looks like this:
The listener take a lot of time

You can see that the calls to the listener and to the `__eq__` method take a lot of time.

After the patch, the calls to `__eq__` are completely gone and the CPU utilization of the `listener` method is down below in the list:

After the patch

dronekit/__init__.py
pietrodn: Micro-benchmark of `None not in` vs. `all(x is not None...)`:

```python
from pymavlink.dialects.v20.common import MAVLink_param_value_message

import timeit


def main():
    list_to_check = [
        MAVLink_param_value_message(param_id='param_{}'.format(i), param_value=1, param_type=4, param_count=501, param_index=i)
        for i in range(500)
    ] + [None]

    assert not all(x is not None for x in list_to_check)

    time_original = timeit.timeit("None not in list_to_check", number=100000, globals=locals())
    time_optimized = timeit.timeit("all(x is not None for x in list_to_check)", number=100000, globals=locals())

    print("Original time: {:.4f}; optimized time: {:.4f}".format(time_original, time_optimized))


if __name__ == '__main__':
    main()
```

Output:

    Original time: 7.4819; optimized time: 3.1870

pietrodn: Once the parameter are loaded, though, the `and` short-circuiting kicks in: `not self._params_loaded` becomes True and the list does not even get checked.

peterbarker: Well, yes, but I find the other change makes things less clear so it would need to be a beneficial change :-)

The reordering is good.

peterbarker:
@pietrodn Needs a rebase.

pietrodn:
Rebased!

peterbarker:
Merged, thanks!

PR #867, opened 2018-10-31
Fix virtualenv installation for CircleCI

pietrodn:
For some reason the CircleCI builds started failing, complaining about a permission error in installing virtualenv at the system level.

It turns out that virtualenv is already bundled in the Docker images: there's no need to install it.


peterbarker:
Merged, thanks!

PR #869, opened 2018-11-01
mavlink.py: specific exception handling for MAVError

pietrodn:
In `mavlink.py`, in `mavlink_thread_in`, specifically catch and ignore `MAVError` (the sporadic messages "invalid MAVLink prefix..." are all MAVErrors), avoiding to swallow legitimate exceptions with a broad `except` clause.

This patch should make it easier to debug DroneKit + pymavlink as library code for developers: it's very hard to diagnose problems if the program silently fails.

### DIFF
--- a/dronekit/mavlink.py
+++ b/dronekit/mavlink.py
@@ -218,11 +218,15 @@ class MAVConnection(object):
                             if error.errno == ECONNABORTED:
                                 raise APIException('Connection aborting during send')
                             raise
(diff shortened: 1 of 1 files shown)


peterbarker:
Thanks, I was fighting with this today.

Incidentally, I don't suppose you know what the currently accepted way of getting a stack trace etc from an exception is?  I think being able to get stuff out of `self._death_error` may be useful.

PR #870, opened 2018-11-01
PEP8 and style improvements

pietrodn:
Ran `flake8` on the code base and fixed most of the issues, with this patch

```
flake8 --ignore=E501,F811,E722,E262,E221 --exclude=dronekit/lib.py dronekit
```

succeeds without errors.
I also implemented some of the suggestions proposed by PyCharm's code inspection tools (chained comparisons and a documentation inconsistency).


dronekit/__init__.py
```diff
@@ -1022,7 +1013,7 @@ def __init__(self, handler):
 
         # Cache all updated attributes for wait_ready.
         # By default, we presume all "commands" are loaded.
-        self._ready_attrs = set(['commands'])
+        self._ready_attrs = {'commands'}
```
peterbarker: This I don't understand.  Moving from a set to a dictionary is strange.

dronekit/__init__.py
```diff
@@ -1345,8 +1335,7 @@ def listener(_):
 
             # Timeouts.
             if self._heartbeat_started:
-                if self._heartbeat_error and self._heartbeat_error > 0 and monotonic.monotonic(
-                ) - self._heartbeat_lastreceived > self._heartbeat_error:
+                if self._heartbeat_error and monotonic.monotonic() - self._heartbeat_lastreceived > self._heartbeat_error > 0:
```
peterbarker: break this to avoid too-long-a-line?

dronekit/__init__.py
```diff
@@ -2084,12 +2071,12 @@ def get_alt():
 
         def check_alt():
             cur = get_alt()
-            delta = abs(alt-cur)
+            delta = abs(alt - cur)
 
             return (
                 (delta < epsilon) or
-                (start < alt and cur > alt) or
```
peterbarker: This obfuscates what's going on, I think.

dronekit/__init__.py
```diff
@@ -3019,6 +3004,8 @@ def connect(ip,
     :returns: A connected vehicle of the type defined in ``vehicle_class`` (a superclass of :py:class:`Vehicle`).
     """
 
+    from dronekit.mavlink import MAVConnection
```
peterbarker: If we're changing this - why are we still polluting the namespace?

dronekit/mavlink.py
```diff
@@ -58,23 +55,23 @@ def __init__(self, device, baud=None, input=True, broadcast=False, source_system
...
                 data, new_addr = self.port.recvfrom(65535)
             except socket.error as e:
-                if e.errno in [ errno.EAGAIN, errno.EWOULDBLOCK, errno.ECONNREFUSED ]:
+                if e.errno in [errno.EAGAIN, errno.EWOULDBLOCK, errno.ECONNREFUSED]:
                     return ""
             if self.udp_server:
                 self.addresses.add(new_addr)
             elif self.broadcast:
-                self.addresses = set([new_addr])
+                self.addresses = {new_addr}
```
peterbarker: Again - moving to dict from set?

dronekit/__init__.py
pietrodn: It’s not a dict. It’s a set literal. https://stackoverflow.com/questions/36674083/why-is-it-possible-to-replace-sometimes-set-with

peterbarker:
Merged ,thanks!

Issue #862, opened 2018-10-25
How to MUTE connection logs?


mustaphos:
Hi.
Is there a way to mute console log outputs?
Thanks for answers.

pietrodn:
You could monkey-patch the methods in `dronekit/util.py`:

```python
from dronekit import util
util.errprinter = lambda msg: None
util.logger = lambda msg: None
```

Caveat: do this before importing the DroneKit module.

pietrodn:
Now dronekit uses the built-in Python `logging` module. Now you can do:

```python
import logging
logging.getLogger('dronekit').addHandler(logging.NullHandler())
logging.getLogger('autopilot').addHandler(logging.NullHandler())
```

Where the `autopilot` logger handles the log messages coming from the firmware, while the `dronekit` logger handles the messages coming from the library.

I think this issue can be closed now.

mustaphos:
@pietrodn thanks for your support.

Issue #118, opened 2015-05-15
Improve DroneKit-Python Debugging


hamishwillee:
DroneKit-Python currently allows debugging using print/log statements and with the command line pdb debugger. It is not particularly easy to use because the console interleaves MAVProxy messages and script print output with the debugging information. In addition, there is no way to launch the script in debug mode - the script has to be modified to include statements that start the debugger once it is running.

We would like to improve the debugging experience. Some ideas include:
- Instructions/information on how to set up debugging using an IDE like Eclipse, PyDev, IDLE
- Provide ability to filter console output display, splitting out debug messages from script output and MAVProxy messages.
- Support for starting a script in a debugger from MAVProxy
- Support to attach to a running script with a debugger and break into it.

Ideas/suggestions/implementations welcome.

tcr3dr:
We can use pdb now. I'm not sure how in-depth to go with this, but `python -m pdb my_dronekit_script.py` lets you debug code straightforwardly.

hamishwillee:
Thanks! I'll have a play with this at some point soon. In the meantime I'll make a small modification to remove any statements related to the old situation.

@tcr3dr One thing, is any logging done automatically? A lot of times on SITL my system is sitting around not arming for no apparent reason. If I was user I'd be a bit clueless on the cause when I got the system back - what would be nice is to be able to get my normal console output. What's the "right thing" to say here.here?

hamishwillee:
#410 fixes this - basically makes it clear that debugging DK-Python is the same as any other debugging task. It uses the example of pdb though. 

Before I close this, would appreciate response about logging - https://github.com/dronekit/dronekit-python/issues/118#issuecomment-153257402 . Maybe we should mention the mav data logs at least.

tcr3dr:
> @tcr3dr One thing, is any logging done automatically? A lot of times on SITL my system is sitting around not arming for no apparent reason. If I was user I'd be a bit clueless on the cause when I got the system back - what would be nice is to be able to get my normal console output. What's the "right thing" to say here?

Are you referring to logging done by MAVProxy in this case? I assumed that is usually what is spit out by SYSTEM_TEXT which we print out by default via `>>> ...`

I'm not too sure what to recommend programmatically, since there are no logs which we can monitor. One handy mechanism would be just watching specific values, which may be a useful abstraction:

```
@vehicle.on_attribute('armed')
def listener(vehicle, name):
    print '>>> Vehicle arm status: %' % vehicle.armed
```

[...]

PR #871, opened 2018-11-04
Logging: use builtin logging module for library and autopilot output

pietrodn:
## Current situation
DroneKit does not play well in the logging department when it's used as a library.
Indeed, the user has no way to configure DroneKit's logging apart from monkey-patching (#862).

The logging of MAVLink `STATUSTEXT` messages can be configured by providing a `status_printer` function to `connect()`; however, while the `STATUSTEXT` message [provides](https://mavlink.io/en/messages/common.html#STATUSTEXT) the severity level ([`MAV_SEVERITY` enum](https://mavlink.io/en/messages/common.html#MAV_SEVERITY)) in a field, DroneKit ignores it and prints all the messages from the autopilot with the same severity.

Also, when an exception is encountered in an attribute or message handler, the stack trace does not get printed, making debugging more difficult.

Related issues: #118, #862

## Proposed changes
This PR tries to improve the situation by replacing all the uses of `errprinter()` and `print()` with the built-in Python [`logging`](https://docs.python.org/3/library/logging.html) module facilities.
The stack trace is printed (`exc_info=True`) in case of "unexpected" exceptions coming from user's code.

`STATUSTEXT` messages are logged by a separate logger to make it clear which messages are coming from the autopilot and which others are emitted from the DroneKit library; the two loggers may be configured independently.
The `STATUSTEXT` messages are printed with the correct logging level specified by the autopilot: filtering by severity is now possible.
The `status_printer` argument of `connect()` is deprecated.

With the proposed changes, the library user would be able to format, filter, redirect and suppress DroneKit's and the autopilot's log messages at his/her pleasure.

## Logging in tests
A nice side effect of this change is that the tests now run completely silent, because nose [captures](https://nose.readthedocs.io/en/latest/plugins/logcapture.html) all the log messages and prints them only in case of error.

To force the printing of log messages in test output, one can do:
```
nosetests -svx --debug=autopilot,dronekit
```

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -25,6 +25,8 @@ Velocity-based movement and control over other vehicle features can be achieved 
 
 It is also possible to work with vehicle "missions" using the :py:attr:`Vehicle.commands` attribute, and run them in AUTO mode.
 
+All the logging is handled through the builtin Python `logging` module.
+
 A number of other useful classes and methods are listed below.
 
 ----
@@ -33,15 +35,13 @@ A number of other useful classes and methods are listed below.
 from __future__ import print_function
 import collections
 import copy
+import logging
 import math
 import struct
-import re
-import sys
 import time
 
 import monotonic
 from past.builtins import basestring
-from dronekit.util import errprinter
 
 from pymavlink import mavutil, mavwp
 from pymavlink.dialects.v10 import ardupilotmega
@@ -531,6 +531,8 @@ class SystemStatus(object):
 
 class HasObservers(object):
     def __init__(self):
+        self._logger = logging.getLogger(__name__)
+
         # A mapping from attr_name to a list of observers
         self._attribute_listeners = {}
         self._attribute_cache = {}
@@ -639,18 +641,14 @@ class HasObservers(object):
         for fn in self._attribute_listeners.get(attr_name, []):
             try:
                 fn(self, attr_name, value)
-            except Exception as e:
-                errprinter('>>> Exception in attribute handler for %s' %
-                           attr_name)
-                errprinter('>>> ' + str(e))
+            except Exception:
+                self._logger.exception('Exception in attribute handler for %s' % attr_name, exc_info=True)
 
         for fn in self._attribute_listeners.get('*', []):
             try:
                 fn(self, attr_name, value)
-            except Exception as e:
-                errprinter('>>> Exception in attribute handler for %s' %
-                           attr_name)
-                errprinter('>>> ' + str(e))
+            except Exception:
+                self._logger.exception('Exception in attribute handler for %s' % attr_name, exc_info=True)
 
     def on_attribute(self, name):
         """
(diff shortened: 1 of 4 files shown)


pietrodn:
In the test output, you may notice that many autopilot logs are at the `CRITICAL` level. This is a bug of ArduPilot [which has already been fixed](https://github.com/ArduPilot/ardupilot/pull/2757). In this PR I decided to respect the MAVLink specification nevertheless.

peterbarker:
I really like the concept.

However, won't the current set of changes here break things for anybody who *is* monkey-patching stuff to clean up the logging?

I mean, we could wave our hands and say "it's master, we're allowed to do that", but given we haven't done a release in... quite some time... that might not be entirely fair.

It would be nice to have some sort of transition arrangement for a year where anybody doing something clever got a warning that errprinter etc was going away.  Can you see a way to do that?

pietrodn:
The current changeset would break all the `errprinter` monkey-patching attempts, in the sense that they would be useless and log messages would be logged to stdout no matter what.

But monkey-patching is a brittle solution anyway, and no one expects libraries not to break when modifying their internals in unexpected ways.

Instead, I think that the biggest retro-compatibility problem is with the `status_printer` argument of the `connect()` function. It allows the user to specify a custom function — accepting a string — to log the *autopilot* (not DroneKit's) messages.
That is a public API and should not break.

In this case, however, we can be smart:
1) if `status_printer` is not specified (default case), we log everything with the new `logging` module: after all, this is what was done before, with print();
2) if `status_printer` is set, we can add a [`Handler`](https://docs.python.org/3/howto/logging.html#handlers) to the autopilot logger, to pass its output through that function, and suppress the default logging to stdout/stderr. DroneKit output would be logged normally (with the logger, to stdout). This strategy should reproduce the current behaviour.

What do you think?

hamishwillee:
@peterbarker FWIW I agree this can reasonably be broken - but we should really make a release to properly document the change (and because it is about time following all recent improvements). Do we have the capacity/ability to make a release or does this require some magic?

peterbarker:
On Sun, 4 Nov 2018, Hamish Willee wrote:

> @peterbarker FWIW I agree this can reasonably be broken - but we should really make a release to properly document the change (and because it is about
> time following all recent improvements). Do we have the capacity/ability to make a release or does this require some magic?

We can make a release happen.

But not on the time scales I'd like to see this change go in :-)

If we can make it compatible, that would be ideal.

pietrodn:
Ok, now the PR does not break the compatibility with old-style `status_printer` arguments via logging handlers.
I also added a test for that.

peterbarker:
Merged, thanks!

"nosetests3" in the root directory yields rather unsatisfactory results, it seems.

pietrodn:
In which sense? This is what I get:

```
(venv)
pietro@ProPietro ~/tmp/dronekit-python
master $ SITL_SPEEDUP=10 nosetests
.............................
----------------------------------------------------------------------
Ran 29 tests in 150.092s

OK
```

PR #875, opened 2018-11-15
mission import / export example : Change file name in message.

YuuyaFujiwara:
I think that it is better the file name to be displayed in the message get from the argument.


peterbarker:
merged, thanks!

YuuyaFujiwara:
You are welcome!
This is my first successful PR.
I'm very happy!

PR #877, opened 2018-11-18
Add commands and tests for sensor calibration

pietrodn:
When handling a drone fleet, it is useful to calibrate the drones from DroneKit's API without resorting to QGroundControl, which requires manual operation.
This is especially true for the calibrations which don't require any manual intervention (gyro, barometer, board level), since they can be carried out on the entire fleet in parallel.

In this PR, I added the support for the calibration of:
1) Magnetometer
2) Accelerometer
3) "Simple accelerometer" calibration
4) Barometer
5) Gyroscope
6) Board level

To test the commands, I developed a generic function able to listen to the `COMMAND_ACK` messages and to assert that a specific command has been ACKed.

The accelerometer calibrations are expected to return the `MAV_RESULT_FAILED` status, since in the SITL there's nobody moving the drone.
Still, it's useful to check the ACK status to make sure that the autopilot does not respond with an "unsupported" or "denied" status.

The magnetometer calibration is expected to return the `MAV_RESULT_UNSUPPORTED` status in the tests because the `MAV_CMD_DO_START_MAG_CAL` is not supported in APM Copter 3.3 (the current version of the SITL), however it is supported in the latest ArduPilot master.
PX4, on the other hand, handles the magnetometer calibration using the standard `MAV_CMD_PREFLIGHT_CALIBRATION` message, so I had to handle the two cases in `Vehicle.calibrate_magnetometer()`.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -2445,7 +2445,6 @@ class Vehicle(HasObservers):
                 0,  # param 7, 1: ESC calibration, 3: barometer temperature calibration
             )
 
-        self._logger.critical(calibration_command)
         self.send_mavlink(calibration_command)
 
     def calibrate_accelerometer(self, simple=False):
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -2427,7 +2427,7 @@ class Vehicle(HasObservers):
                 1,  # param 2, Automatically retry on failure (0=no retry, 1=retry).
                 1,  # param 3, Save without user input (0=require input, 1=autosave).
                 0,  # param 4, Delay (seconds).
-                1,  # param 5, Autoreboot (0=user reboot, 1=autoreboot).
+                0,  # param 5, Autoreboot (0=user reboot, 1=autoreboot).
                 0,  # param 6, Empty.
                 0,  # param 7, Empty.
             )
@@ -2468,8 +2468,8 @@ class Vehicle(HasObservers):
         )
         self.send_mavlink(calibration_command)
 
-    def calibrate_board_level(self):
-        """Request board level calibration."""
+    def calibrate_vehicle_level(self):
+        """Request vehicle level (accelerometer trim) calibration."""
 
         calibration_command = self.message_factory.command_long_encode(
             self._handler.target_system, 0,  # target_system, target_component
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -2397,6 +2397,125 @@ class Vehicle(HasObservers):
 
         self.send_mavlink(reboot_msg)
 
+    def calibrate_gyro(self):
+        """Request gyroscope calibration."""
+
+        calibration_command = self.message_factory.command_long_encode(
+            0, 0,  # target_system, target_component
+            mavutil.mavlink.MAV_CMD_PREFLIGHT_CALIBRATION,  # command
+            0,  # confirmation
+            1,  # param 1, 1: gyro calibration, 3: gyro temperature calibration
+            0,  # param 2, 1: magnetometer calibration
+            0,  # param 3, 1: ground pressure calibration
+            0,  # param 4, 1: radio RC calibration, 2: RC trim calibration
+            0,  # param 5, 1: accelerometer calibration, 2: board level calibration, 3: accelerometer temperature calibration, 4: simple accelerometer calibration
+            0,  # param 6, 2: airspeed calibration
+            0,  # param 7, 1: ESC calibration, 3: barometer temperature calibration
+        )
+        self.send_mavlink(calibration_command)
+
+    def calibrate_magnetometer(self):
+        """Request magnetometer calibration."""
+
+        # APM requires the MAV_CMD_DO_START_MAG_CAL command, only present in the APM MAVLink dialect
+        if self._autopilot_type == mavutil.mavlink.MAV_AUTOPILOT_ARDUPILOTMEGA:
+            calibration_command = self.message_factory.command_long_encode(
+                0, 0,  # target_system, target_component
+                mavutil.mavlink.MAV_CMD_DO_START_MAG_CAL,  # command
+                0,  # confirmation
+                0,  # param 1, uint8_t bitmask of magnetometers (0 means all).
+                1,  # param 2, Automatically retry on failure (0=no retry, 1=retry).
+                1,  # param 3, Save without user input (0=require input, 1=autosave).
+                0,  # param 4, Delay (seconds).
+                1,  # param 5, Autoreboot (0=user reboot, 1=autoreboot).
+                0,  # param 6, Empty.
+                0,  # param 7, Empty.
+            )
+        else:
+            calibration_command = self.message_factory.command_long_encode(
+                0, 0,  # target_system, target_component
+                mavutil.mavlink.MAV_CMD_PREFLIGHT_CALIBRATION,  # command
+                0,  # confirmation
+                0,  # param 1, 1: gyro calibration, 3: gyro temperature calibration
+                1,  # param 2, 1: magnetometer calibration
+                0,  # param 3, 1: ground pressure calibration
+                0,  # param 4, 1: radio RC calibration, 2: RC trim calibration
+                0,  # param 5, 1: accelerometer calibration, 2: board level calibration, 3: accelerometer temperature calibration, 4: simple accelerometer calibration
+                0,  # param 6, 2: airspeed calibration
+                0,  # param 7, 1: ESC calibration, 3: barometer temperature calibration
+            )
+
+        self._logger.critical(calibration_command)
+        self.send_mavlink(calibration_command)
+
+    def calibrate_accelerometer(self):
+        """Request accelerometer calibration."""
+
+        calibration_command = self.message_factory.command_long_encode(
+            0, 0,  # target_system, target_component
+            mavutil.mavlink.MAV_CMD_PREFLIGHT_CALIBRATION,  # command
+            0,  # confirmation
+            0,  # param 1, 1: gyro calibration, 3: gyro temperature calibration
+            0,  # param 2, 1: magnetometer calibration
+            0,  # param 3, 1: ground pressure calibration
+            0,  # param 4, 1: radio RC calibration, 2: RC trim calibration
+            1,  # param 5, 1: accelerometer calibration, 2: board level calibration, 3: accelerometer temperature calibration, 4: simple accelerometer calibration
+            0,  # param 6, 2: airspeed calibration
+            0,  # param 7, 1: ESC calibration, 3: barometer temperature calibration
+        )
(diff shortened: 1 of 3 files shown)


hamishwillee:
Looks great. @peterbarker Is the right person to review this.

Just FMI, does a calibration "clear" existing settings when started - ie might you choose not to run an accelerometer calibration, because you could be clearing an already properly calibrated vehicle?

dronekit/__init__.py
```diff
@@ -2397,6 +2397,125 @@ def reboot(self):
...
+            0,  # param 5, 1: accelerometer calibration, 2: board level calibration, 3: accelerometer temperature calibration, 4: simple accelerometer calibration
+            0,  # param 6, 2: airspeed calibration
+            0,  # param 7, 1: ESC calibration, 3: barometer temperature calibration
+        )
+        self.send_mavlink(calibration_command)
+
+    def calibrate_magnetometer(self):
+        """Request magnetometer calibration."""
+
+        # APM requires the MAV_CMD_DO_START_MAG_CAL command, only present in the APM MAVLink dialect
```
peterbarker: ArduPilot, not APM, please :-)

peterbarker: Also, probably better to specify `ardupilotmega.xml`

dronekit/__init__.py
```diff
@@ -2397,6 +2397,125 @@ def reboot(self):
...
+        if self._autopilot_type == mavutil.mavlink.MAV_AUTOPILOT_ARDUPILOTMEGA:
+            calibration_command = self.message_factory.command_long_encode(
+                0, 0,  # target_system, target_component
+                mavutil.mavlink.MAV_CMD_DO_START_MAG_CAL,  # command
+                0,  # confirmation
+                0,  # param 1, uint8_t bitmask of magnetometers (0 means all).
+                1,  # param 2, Automatically retry on failure (0=no retry, 1=retry).
+                1,  # param 3, Save without user input (0=require input, 1=autosave).
+                0,  # param 4, Delay (seconds).
+                1,  # param 5, Autoreboot (0=user reboot, 1=autoreboot).
```
peterbarker: I think I'd like to see a parameter for this one.

We'd be looking for consistency across vehicles here, I think; ArduPilot should be left in the same state as the PX4 firmware; either requiring a reboot or not, depending on whetherPX4 does or not after a mag cal.

pietrodn:
Yes, if you calibrate a vehicle the previous settings (e.g. accelerometer offset and scale parameters) are cleared and replaced.

dronekit/__init__.py
```diff
@@ -2397,6 +2397,125 @@ def reboot(self):
...
+        )
+        self.send_mavlink(calibration_command)
+
+    def calibrate_magnetometer(self):
+        """Request magnetometer calibration."""
+
+        # APM requires the MAV_CMD_DO_START_MAG_CAL command, only present in the APM MAVLink dialect
+        if self._autopilot_type == mavutil.mavlink.MAV_AUTOPILOT_ARDUPILOTMEGA:
+            calibration_command = self.message_factory.command_long_encode(
+                0, 0,  # target_system, target_component
```
peterbarker: This looks wrong.  You should be targetting a specify vehicle.

Same is true of the other messages.

dronekit/__init__.py
```diff
@@ -2397,6 +2397,125 @@ def reboot(self):
...
+            0,  # param 2, 1: magnetometer calibration
+            0,  # param 3, 1: ground pressure calibration
+            0,  # param 4, 1: radio RC calibration, 2: RC trim calibration
+            1,  # param 5, 1: accelerometer calibration, 2: board level calibration, 3: accelerometer temperature calibration, 4: simple accelerometer calibration
+            0,  # param 6, 2: airspeed calibration
+            0,  # param 7, 1: ESC calibration, 3: barometer temperature calibration
+        )
+        self.send_mavlink(calibration_command)
+
+    def calibrate_accelerometer_simple(self):
```
peterbarker: Perhaps have a "simple=True" parameter instead of a separate function?

dronekit/__init__.py
```diff
@@ -2397,6 +2397,125 @@ def reboot(self):
...
+            0,  # param 2, 1: magnetometer calibration
+            0,  # param 3, 1: ground pressure calibration
+            0,  # param 4, 1: radio RC calibration, 2: RC trim calibration
+            4,  # param 5, 1: accelerometer calibration, 2: board level calibration, 3: accelerometer temperature calibration, 4: simple accelerometer calibration
+            0,  # param 6, 2: airspeed calibration
+            0,  # param 7, 1: ESC calibration, 3: barometer temperature calibration
+        )
+        self.send_mavlink(calibration_command)
+
+    def calibrate_board_level(self):
```
peterbarker: My understanding is this is usually used for "trimming" the board; the accelerometer calibration is really setting "level" on a gross scale.  Perhaps we can find a more illuminating name here?  `trim_board_level`?

hamishwillee:
> Yes, if you calibrate a vehicle the previous settings (e.g. accelerometer offset and scale parameters) are cleared and replaced.

Thanks. So arguably you might choose not to do this "automatically" for acceleration because you wouldn't want to undo "possibly good" settings (at least without checking the current calibration state).

dronekit/__init__.py
pietrodn: I specified `self._handler.target_system` as the target system for all commands.

dronekit/__init__.py
pietrodn: Done.

dronekit/__init__.py
pietrodn: On QGroundControl for PX4, the calibration is called "Level horizon". I'd like to keep the "calibrate" in the method name.

peterbarker:
On Sun, 18 Nov 2018, Pietro De Nicolao wrote:

> +
> +    def calibrate_board_level(self):
> 
> On QGroundControl for PX4, the calibration is called "Level horizon". I'd
> like to keep the "calibrate" in the method name.

You know the board may not be even close to level at that point, 'though?

ArduPilot supports dozens of board orientations - and I very much doubt 
PX4 is different.

pietrodn:
From what I know, that calibration tells the board which orientation corresponds to 0° roll, 0° pitch. Isn't it the same of saying "board level" (or better, "vehicle level")?

peterbarker:
On Tue, 20 Nov 2018, Pietro De Nicolao wrote:

> From what I know, that calibration tells the board which orientation
> corresponds to 0° roll, 0° pitch. Isn't it the same of saying "board level"
> (or better, "vehicle level")?

vehicle level should be appropriate (give or take tailsitters on some 
autopilot systems, probably!).

Board level is *not*.  Some indication: 
https://github.com/ardupilot/ardupilot/blob/master/libraries/AP_Math/vector3.cpp#L35

Several of my vehicles have their autopilots oriented upside down, for 
example.

dronekit/__init__.py
pietrodn: PX4 does not reboot automatically after a mag calibration.
Some people say that the reboot isn't sufficient and recommend a full power cycle, however: https://github.com/PX4/Firmware/issues/9030

pietrodn: Ok, I disabled auto-reboot for mag calibration on ArduPilot, so that the behaviour is consistent with PX4.
It is always possible to reboot using the `Vehicle.reboot()` method.

dronekit/__init__.py
pietrodn: I renamed the method `calibrate_vehicle_level`.

pietrodn:
> The functions here are a bit of a mis-mash, really. Some of the functions are done straight-away (e.g. the AHRS trimming, simple accel cal). Some of them require significant user interaction (e.g. the mag calibration.) Not saying they're not a good idea, just that there are multiple concepts/workflows involved in these and it might be worth considering that.

@peterbarker I think I've resolved all the other issues; how do we deal with this?
My user experience with API-initiated calibration is the following:

1) For the calibrations which do not require user interaction (gyro, vehicle level), I select all the drones in my application and run the calibration in parallel on all of them. `STATUSTEXT` messages and the beeping of the boards indicate the progress of the calibration.
2) For the calibrations which do require user interaction (accelerometer, magnetometer) I calibrate the drones one by one. This is still convenient as I don't have to open and connect QGroundControl to each drone, which is a time-consuming process. In this case I still rely on the aid of `STATUSTEXT` messages and beeps to know how to move the vehicle.

In this workflow, I think it's reasonable to have API methods to *initiate* the calibration. AFAIK there's no way, in MAVLink, to read the calibration progress other than parsing `STATUSTEXT` messages.

peterbarker:
On Sat, 24 Nov 2018, Pietro De Nicolao wrote:

>  1. For the calibrations which do not require user interaction (gyro, vehicle level), I select all the drones in my application and run the calibration in
>     parallel on all of them. STATUSTEXT messages and the beeping of the boards indicate the progress of the calibration.

Just musing: the SkyViper production line had a *very* simple calibration 
procedure which didn't require anything but putting the vehicle in a clamp 
and pressing "go".  So it *is* possible to do things very smoothly :-)

>  2. For the calibrations which do require user interaction (accelerometer, magnetometer) I calibrate the drones one by one. This is still convenient as I
>     don't have to open and connect QGroundControl to each drone, which is a time-consuming process. In this case I still rely on the aid of STATUSTEXT
>     messages and beeps to know how to move the vehicle.

> In this workflow, I think it's reasonable to have API methods to initiate the calibration. AFAIK there's no way, in MAVLink, to read the calibration
> progress other than parsing STATUSTEXT messages.

I'm not familiar with PX4's calibration procedures, sorry :-)

In fact, it occurs to me that the magnetometer calibration in this PR is 
not the way ArduPilot does things.  It's all on-board with 
MAG_CAL_PROGRESS messages informing the GCS what's going on.

pietrodn:
This PR has no support for tracking calibration progress, still the `MAV_CMD_DO_START_MAG_CAL` command looks like the correct way to initiate a magnetometer calibration on ArduPilot: https://github.com/ArduPilot/ardupilot/blob/027beb0dc1feb59d12e71ffeb4d9d07c09e37bd6/libraries/AP_Compass/AP_Compass_Calibration.cpp#L290

pietrodn:
I see that in the [ardupilotmega.xml](https://mavlink.io/en/messages/ardupilotmega.html) dialect of MAVLink there are some commands specific to magnetometer calibration.
However, I have no idea how to correctly use them.

Do you think this PR is viable without the magnetometer calibration part?

hamishwillee:
@peterbarker Do you have an opinion on this?
My general answer to this sort of thing is "push it in as-is and revisit the other bits when you can get clarity".

peterbarker:
Looking good.  I think I'll merge this as-is, 'though I'm not sure we're choosing the correct calibration routines for ArduPilot.

PR #892, opened 2018-12-28
add logging basic config

Huibean:
fix error with
```shell
  No handlers could be found for logger "dronekit"
```

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -532,6 +532,7 @@ class SystemStatus(object):
 
 class HasObservers(object):
     def __init__(self):
+        logging.basicConfig()
         self._logger = logging.getLogger(__name__)
 
         # A mapping from attr_name to a list of observers


pietrodn:
The issue occurs in Python 2.7, but not in Python 3.7.
In Python 3.7, by default the records with level WARNING or more [are logged to the console](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library).
The issue does not appear in tests, because nosetests also defines its own handlers.

The problem with `logging.basicConfig()` is that you can call it only once: after the first invocation, all the other are ignored.
So the user would lose the ability to invoke `basicConfig()` after `import dronekit`.

I think the ideal solution would be to add a `basicConfig()` invocation with decent formatting after the import block (e.g. [at line 50 in `__init__.py`](https://github.com/dronekit/dronekit-python/blob/master/dronekit/__init__.py#L50)):

```python
logging.basicConfig(level=logging.INFO, format="[%(asctime)s] %(levelname)s: %(name)s: %(message)s")
```

In this way if the user wants to configure their own logging, they can do it before importing `dronekit` and our `basicConfig()` would be ignored; otherwise, if no such configuration is defined, we end up with good defaults.

pietrodn:
_[The Hitchhiker's Guide to Python](https://docs.python-guide.org/writing/logging/)_ suggests to add only NullHandlers to the root logger, but in this way we would lose all log messages by default, and that would be confusing for new users.

PR #896, opened 2019-01-08
add-status-text-handler

Huibean:
add this feature so that user can access status text message outside dronekit package by defining their statustext_handler

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1033,6 +1033,7 @@ class Vehicle(HasObservers):
 
         self._handler = handler
         self._master = handler.master
+        self.statustext_handler = None
 
         # Cache all updated attributes for wait_ready.
         # By default, we presume all "commands" are loaded.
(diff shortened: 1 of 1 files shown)


pietrodn:
I'm not sure this is the best way to handle `STATUSTEXT` messages externally.

One could still do:
```python
import logging
statustext_logger = logging.getLogger('autopilot')
statustext_logger.addHandler(MyHandler())
```

using the facilities provided in the builtin Python [`logging`](https://docs.python.org/3/library/logging.html) module.

The `dronekit.connect()` function does exactly that ([here](https://github.com/dronekit/dronekit-python/blob/master/dronekit/__init__.py#L3149)). For backwards compatibility, the user can pass a custom function as the `status_printer` parameter. If set, that function will be called with the `STATUSTEXT` messages.

peterbarker:
@Huibean Any reason not to add the decorator to your own function?  Why hook it this way?

e.g. https://github.com/dronekit/dronekit-python/blob/master/examples/create_attribute/my_vehicle.py#L63

Huibean:
@peterbarker create sub vehicle class can handle this case， but there would be two listeners for same mavlink message

peterbarker:
On Thu, 10 Jan 2019, Huibean Luo wrote:

> @peterbarker create sub vehicle class can handle this case， but there would be two listeners for same mavlink message

Yep, that's fine.  It's a list of functions per message...

Huibean:
@peterbarker so, I could close this now

PR #904, opened 2019-01-31
Allowing specification of timeout for uploading missions

afsafzal:
Without a timeout, uploading missions can get stuck in an infinite loop if a buggy version of vehicle refuses to accept the mission. With a timeout parameter, the infinite loop can be avoided.


peterbarker:
Merged, thanks

PR #906, opened 2019-02-08
Add Source Component ID parameter 

jmachuca77:
Added a source component ID to connect function to allow setting a different comp ID than 0. Default to 0 to make it backwards compatible.


peterbarker:
Merged, thanks.

The source component stuff hasn't been in pymavlink all that long, so you get to handle the "unknown named parameter" support questions :-P

PR #910, opened 2019-02-13
fix URL to mavproxy documentation

drnic:
I think this link is now supposed to go to http://ardupilot.github.io/MAVProxy/html/getting_started/starting.html

- [ ] Um, I'm unsure why all lines are modified. Fixing.


drnic:
Hmm, I think this file has different line endings from VS Code/OS X and so its editing all lines. I'll close this PR and try to submit new PR via github.com directly.

PR #916, opened 2019-03-05
Adding README to PyPi & minimizing imports

MohamedAliRashad:
Adding the README as a long description in PyPi and minimizing the ```setuptools``` import to only ```setup``` and ```find_packages```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,20 @@
-import setuptools
+from setuptools import find_packages, setup
 
 version = '2.9.1'
 
-setuptools.setup(
+try:
+    with open(os.path.join(here, 'README.md'), encoding='utf-8') as f:
+        LongDescription = f.read()
+except FileNotFoundError:
+    LongDescription = 'Python API for communication and control of drones over MAVLink.'
+
+
+setup(
     name='dronekit',
     zip_safe=True,
     version=version,
     description='Developer Tools for Drones.',
-    long_description='Python API for communication and control of drones over MAVLink.',
+    long_description=LongDescription,
     url='https://github.com/dronekit/dronekit-python',
     author='3D Robotics',
     install_requires=[
@@ -26,5 +33,5 @@ setuptools.setup(
         'Topic :: Scientific/Engineering',
     ],
     license='apache',
-    packages=setuptools.find_packages()
+    packages=find_packages()
 )
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,12 @@
 from setuptools import find_packages, setup
+import os
 
 version = '2.9.1'
 
 try:
-    with open(os.path.join(here, 'README.md'), encoding='utf-8') as f:
+    with open(os.path.join(os.path.dirname(__file__), 'README.md'), encoding='utf-8') as f:
         LongDescription = f.read()
-except FileNotFoundError:
+except:
     LongDescription = 'Python API for communication and control of drones over MAVLink.'
 
 


setup.py
```diff
@@ -1,13 +1,21 @@
-import setuptools
+from setuptools import find_packages, setup
```
peterbarker: Is there any particular reason to pull these into our namespace?

setup.py
```diff
@@ -1,13 +1,21 @@
...
+import os
 
 version = '2.9.1'
 
-setuptools.setup(
+try:
+    with open(os.path.join(os.path.dirname(__file__), 'README.md'), encoding='utf-8') as f:
```
peterbarker: Is this sort of thing typical on PyPi?

setup.py
```diff
@@ -26,5 +34,5 @@
         'Topic :: Scientific/Engineering',
     ],
     license='apache',
-    packages=setuptools.find_packages()
```
peterbarker: this change can be reverted if we don't specify symbols to import...

setup.py
```diff
@@ -1,13 +1,21 @@
-import setuptools
+from setuptools import find_packages, setup
+import os
 
 version = '2.9.1'
 
-setuptools.setup(
```
amilcarlucas: This one can also be reverted if you do not pull setup into the namespace.

setup.py
MohamedAliRashad: well, i was going through the ```setup.py``` file of some open source repositerioes and i found that a lot of the big names in software uses ```from ... import ...``` instead of ```import``` so i thought it would be nice to add it here.

But to be true when i digged more into this i found no real gain for this solution, so if you don't prefer it i have no problem taking it down.

setup.py
MohamedAliRashad: yeah, as i said to @peterbarker i have no problem pulling this down if you didn't like it.

setup.py
MohamedAliRashad: i found it in the Django ```setup.py``` [here](https://github.com/django/django/blob/782d85b6dfa191e67c0f1d572641d8236c79174c/setup.py#L64)

setup.py
MohamedAliRashad: the importing style should not cause much problem i think.

MohamedAliRashad:
@peterbarker i just saw a big project like [Django](https://github.com/django/django/blob/782d85b6dfa191e67c0f1d572641d8236c79174c/setup.py#L5) uses it, so i thought it would be useful here ... i am very open to closing this one out and opening a new one with only the README in.

peterbarker:
On Wed, 13 Mar 2019, Mohamed Rashad wrote:

> @peterbarker i just saw a big project like Django uses it, so i thought it would be useful here ... i am very open to closing this one out and opening a
> new one with only the README in.

No need to open a PR :-)

Make a backup branch: git branch patch-2-201903140908 # happy pi day!
revert the three commits, keeping the code: git reset --soft HEAD~3
unstage the deltas you don't want to keep: git reset -p
commit the deltas you do want to keep: git commit -m "your commit message 
here"
get rid of the deltas you don't want: git checkout -p

force push the branch up to github: git push -f github patch-2  # assuming 
github is your remote name for upstream; could be "origin"

setup.py
peterbarker: Please do.

peterbarker:
@MohamedAliRashad No pressure - but I'd like to get this in before doing the release.

If you can't get to it I'd be quite happy to make the discussed changes.

Please let me know :-)

MohamedAliRashad:
I'm sorry for this mess but i decided to give GitKraken a shot so i can be reliefed from all of this headache about git squashing, rebasing and merging ... but it seems i am screwing stuff even more ^ ^

MohamedAliRashad:
@peterbarker I give up, There is some sort of permission denial that i can't get over it ... please merge the changes as you like, i hope i have helped.

peterbarker:
@MohamedAliRashad Was the permission denial a "non-fast-forward" error?  If so, simply means you have to "force" the push up to github.  I don't know how you do that in gitkraken, on the commandline it's the "-f" flag.

MohamedAliRashad:
@peterbarker No, it's not it. 
i am suspecting that this error is due to the changes i made to my account after i pulled this request, GitHub is not accepting me as the user i guess ... Anyway i just knew that you were able to squash my commits all this time but left me to suffer and learn ^ ^
image

amilcarlucas:
@peterbarker this can be closed, right ?

peterbarker:
Thanks!

PR #920, opened 2019-03-16
Correct mask to be able to control attitude

baptistelabat:
Pitch desired was not changing with previous mask. Tested with bebop2

### DIFF
--- a/examples/set_attitude_target/set_attitude_target.py
+++ b/examples/set_attitude_target/set_attitude_target.py
@@ -103,7 +103,7 @@ def send_attitude_target(roll_angle = 0.0, pitch_angle = 0.0,
         0, # time_boot_ms
         1, # Target system
         1, # Target component
-        0b00000000 if use_yaw_rate else 0b00000100,
+        0b00000011 if use_yaw_rate else 0b00000111,
         to_quaternion(roll_angle, pitch_angle, yaw_angle), # Quaternion
         0, # Body roll rate in radian
         0, # Body pitch rate in radian


peterbarker:
I'm confused.  These are "ignore" bits - indicating which axes the vehicle should NOT honour in the message.

This PR seems to be telling the vehicle to ignore roll/pitch - the opposite of the title.

What were you seeing before the patch?

What are you seeing after the patch?

Which firmware is running on the vehicle?

peterbarker:
(definitely something I'd like to get sorted before doing a release!)

baptistelabat:
From what I understood looking at other messages (position and speed), the last two bits correspond to roll and pitch rate, not roll and pitch. In the example roll and pitch rate are set to zero, which I want to ignore.
I was using firmware 3.5.7 (a bit old) and testing without propeller, moving the drone by hand for safety issues for now.
Before the patch, the motors were just turning at the same speed when holding position and when requested to move forward (negative pitch) then backward. I was able to check in the log, that the requested pitch was not changing, which I assumed was due to the confict with 0 pitch rate requested.
After the patch, the motors were changing speed, and I could check in the log that the requested pitch was changing.
I updated today to 3.6.7 to test again and got different results : when I first tested with the version before the patch (00000003.BIN), it worked as expected, which surprised me. However, I repeated the test several times and did not work again. When tested with version after patch it worked every time I tested (00000004.BIN)

[0000000.zip](https://github.com/dronekit/dronekit-python/files/2979429/0000000.zip)

PR #949, opened 2019-06-25
Change the name of the mavlink message.

roman-dvorak:
In the latest version, the reception of RC channels with PX4 autopilot did not work. This commit repairs bad mavlink message name and adds support for more (18) channels. Tested with PX4

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1142,31 +1142,18 @@ class Vehicle(HasObservers):
         # All keys are strings.
         self._channels = Channels(self, 8)
 
+        @self.on_message('RC_CHANNELS_RAW')
         @self.on_message('RC_CHANNELS')
         def listener(self, name, m):
             def set_rc(chnum, v):
                 '''Private utility for handling rc channel messages'''
                 # use port to allow ch nums greater than 8
-                self._channels._update_channel(str(0 * 8 + chnum), v)
+                port = 0 if name == "RC_CHANNELS" else m.port
+                self._channels._update_channel(str(port * 8 + chnum), v)
 
-            set_rc(1, m.chan1_raw)
-            set_rc(2, m.chan2_raw)
-            set_rc(3, m.chan3_raw)
-            set_rc(4, m.chan4_raw)
-            set_rc(5, m.chan5_raw)
-            set_rc(6, m.chan6_raw)
-            set_rc(7, m.chan7_raw)
-            set_rc(8, m.chan8_raw)
-            set_rc(9, m.chan9_raw)
-            set_rc(10, m.chan10_raw)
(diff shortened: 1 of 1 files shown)


dronekit/__init__.py
```diff
@@ -1142,12 +1142,12 @@ def listener(vehicle, name, m):
...
         self._channels = Channels(self, 8)
 
-        @self.on_message('RC_CHANNELS_RAW')
+        @self.on_message('RC_CHANNELS')
         def listener(self, name, m):
             def set_rc(chnum, v):
                 '''Private utility for handling rc channel messages'''
                 # use port to allow ch nums greater than 8
-                self._channels._update_channel(str(m.port * 8 + chnum), v)
+                self._channels._update_channel(str(0 * 8 + chnum), v)
```
peterbarker: What about simply `chnum` :-)

I'm hoping your testing will show that casting this to a string is unnecessary.

dronekit/__init__.py
```diff
@@ -1157,6 +1157,16 @@ def set_rc(chnum, v):
             set_rc(6, m.chan6_raw)
             set_rc(7, m.chan7_raw)
             set_rc(8, m.chan8_raw)
+            set_rc(9, m.chan9_raw)
```
peterbarker: This could be done with a loop and `set_rc(i, m.getattr("chan%u_raw", i))` - but fine as-is.

dronekit/__init__.py
roman-dvorak: Sure!

roman-dvorak:
Dear @peterbarker, I fixed the above bugs and added backward compatibility for old messages.

roman-dvorak:
the new message also provides RSSI. I'm wondering if i should also add RSSI to this field to channels dictionary.

peterbarker:
@roman-dvorak tests need to pass.

You can run the failing one with `nosetests -v -v  ./dronekit/test/sitl/test_channels.py`

This patch seems to be required:
```
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -1142,8 +1142,7 @@ class Vehicle(HasObservers):
         # All keys are strings.
         self._channels = Channels(self, 8)
 
-        @self.on_message('RC_CHANNELS_RAW')
-        @self.on_message('RC_CHANNELS')
+        @self.on_message(['RC_CHANNELS', 'RC_CHANNELS_RAW'])
         def listener(self, name, m):
```

... and past that there's an assertion on the number of channels...

roman-dvorak:
@peterbarker  Are any further edits required for this pull request?

peterbarker:
@roman-dvorak Merged, thanks!

PR #1002, opened 2020-02-16
Fixed gimbal target location

thorstenbar:
The Gimbal class is not targeting absolute gps coordinates. This fixes the issue.


kiranshila:
I submitted a PR for this in November, which fixes this among other bugs with the gimbal class, but the package maintainers don't seem to be working on dronekit anymore

minhaj6:
this project is basically abandoned at this point. this is unfortunate.

PR #1023, opened 2020-04-23
fixed last_heartbeat function

tridge:
this should return the time since the last heartbeat, but instead it
was giving the time between the last hearbeat and the last mavlink
pkt. That means if you lose link completely then last_heartbeat
wouldn't update


tridge:
testing passed

kmk142789 (approved):

PR #1093, opened 2021-02-16
Wind stats addition

o-gent:
Pretty simple addition to add wind stats to the vehicle class. Tried to follow convention seen in other attribute classes and methods

Needed it for a project I'm doing so maybe it could be useful for other people!

Tested with an STL and is functional

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -220,6 +220,25 @@ class GPSInfo(object):
         return "GPSInfo:fix=%s,num_sat=%s" % (self.fix_type, self.satellites_visible)
 
 
+class Wind(object):
+    """
+    Wind information
+
+    An object of this type is returned by :py:attr: `Vehicle.wind`.
+
+    :param wind_direction: Wind direction in degrees
+    :param wind_speed: Wind speed in m/s
+    :param wind_speed_z: vertical wind speed in m/s
+    """
+    def __init__(self, wind_direction, wind_speed, wind_speed_z):
+        self.wind_direction = wind_direction
+        self.wind_speed = wind_speed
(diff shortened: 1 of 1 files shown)


o-gent:
Those tests look pretty messed up, probably worth just removing any failing tests - looks like they've been failing for a long time. test_goto.py is fairly poor quality anyway but fails when trying to change mode for some reason.. Create tasks for code to be tested. I tried to get nosetests set up but wouldn't run anything even though I'm using the same command as travis-ci

o-gent:
The tests that are failing (from wayyy before my commit) are actually working on my SITL. Looks like dronekit-sitl isn't working. Could make the tests use the manually installed SITL then don't need to fix dronekit-sitl?

o-gent:
Hi @peterbarker, is there anything else I can do to get this integrated?

dronekit/__init__.py
```diff
@@ -1678,6 +1710,13 @@ def listener(self, attr_name, value):
         """
         return self._location
 
+    @property
+    def wind(self):
+        """
+        Current wind status (:pu:class: `Wind`)
+        """
+        return Wind(self._wind_direction, self._wind_speed, self._wind_speed_z)
```
peterbarker, line 1718: Shouldn't this, like battery, return `None` if the attributes are all `None`?

o-gent: Done

peterbarker:
@o-gent new changes have been tested?

o-gent:
@peterbarker I've tested it on my STL
image
but I haven't written a test as it seems the STL used in testing isn't working the same as what I see in my STL

peterbarker:
Thanks @o-gent , merging.

PR #1140, opened 2021-12-16
Correct import of MutableMapping for Python 3.10

peterbarker:


KrishnanS2006:
This is exactly the change I'm looking for (I've manually been changing the `collections.MutableMapping` to `collections.abc.MutableMapping` every time) - but this change never seems to be reflected when pulling from `pypi`. Any ideas why? It'd be very convenient not to manually edit the virtual environment's site-packages every time I install the package.

PR #1190, opened 2023-06-08
Maintainers Needed

mrpollo:
Hey @hamishwillee, your feedback would be appreciated.


hamishwillee (approved):
This provides a clear indication that we want more maintainers, and how they can apply. I think it is good and friendly.

It might not satisfy the "need" or the original posters, but as I said there, it is not clear how valid that is.

PR #1198, opened 2023-07-12
Get RAW values of the servo outputs

TongBingda:
@morzack Thank you for replying to my email, I would love to be a maintainer of this program!

The RAW value of the servo output can be acquired based on the SERVO_OUTPUT_RAW information.

An object of this type is returned by :py:attr:`Vehicle.servos`.

:param port: Servo output port (set of 8 outputs = 1 port). Flight stacks running on Pixhawk should use: 0 = MAIN, 1 = AUX.
:param servo_raw['1'] ~ servo_raw['16']: Servo output 1 ~ 16 value.

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -247,6 +247,23 @@ class Wind(object):
         return "Wind: wind direction: {}, wind speed: {}, wind speed z: {}".format(self.wind_direction, self.wind_speed, self.wind_speed_z)
 
 
+class Servos(object):
+    """
+    The RAW values of the servo outputs.
+
+    An object of this type is returned by :py:attr:`Vehicle.servos`.
+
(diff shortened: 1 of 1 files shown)


morzack:
Sorry for the delay in taking a look at this. Overall, it looks good to me. Did you do any testing on actual hardware for this change?
@TongBingda

TongBingda:
> Sorry for the delay in taking a look at this. Overall, it looks good to me. Did you do any testing on actual hardware for this change? @TongBingda

Yes, I recently tested this new feature using a Cube Orange and a CUAV X7+. I found some problems with the previous code. When running ardupilot firmware using the sitl method, SERVO_OUTPUT_RAW usually uses mavlink 2.0, whereas this message usually uses mavlink 1.0 when I run ardupilot firmware using Cube Orange, which makes it impossible to get the servo9_raw to servo16 _raw values. So in the latest code I used len(m.lengths) to get the actual number of servo_outputs to prevent errors.

PR #1204, opened 2023-09-22
Update __init__.py-line no: 2920, class Command argument message type…

RedWvn:
The Command class in __init__.py uses mavutil.mavlink.MAVLink_mission_item_message for the message type, which uses the less accurate float32 for sending the long and lat parameters. This can be made more accurate if we use the mavutil.mavlink.MAVLink_mission_item_int_message type which uses int32 for sending, and is the recommended implementation. Hence, for the Command class, we have changed the message type to message_item_int since message_item is now depreciated.


hamishwillee:
Thanks. This looks great, but same fix in #1209 is more complete.

PR #1238, opened 2025-06-15
Fix compatibility issue with Python 3.10 by updating MutableMapping import

dododoyo:
The `collections.MutableMapping` class was deprecated in Python 3.3 and removed in Python 3.10. This caused an `AttributeError` in DroneKit when running under Python 3.10. Updated the `Parameters` class in `dronekit/__init__.py` to inherit from `collections.abc.MutableMapping` instead of `collections.MutableMapping`. This change ensures compatibility with Python 3.10 and later while maintaining support for older Python

### DIFF
--- a/dronekit/__init__.py
+++ b/dronekit/__init__.py
@@ -36,8 +36,9 @@ import sys
 import collections
 
 # Python3.10 removed MutableMapping from collections:
-if sys.version_info.major == 3 and sys.version_info.minor >= 10:
-    from collections.abc import MutableMapping
+if sys.version_info.major == 3:
+    if sys.version_info.minor >= 10:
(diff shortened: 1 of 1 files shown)


dronekit/__init__.py
```diff
@@ -36,8 +36,9 @@
 import collections
 
 # Python3.10 removed MutableMapping from collections:
-if sys.version_info.major == 3 and sys.version_info.minor >= 10:
-    from collections.abc import MutableMapping
+if sys.version_info.major == 3:
+    if sys.version_info.minor >= 10:
+        from collections.abc import MutableMapping
```
hamishwillee, line 41: This is effectively identical

PR #1253, opened 2026-08-25
Remove undeclared dependency on past.builtins

### LINKED ISSUE #1252: ModuleNotFoundError: No module named 'past'
In line 51 of `__init__.py`, there is the following line:
```python
from past.builtins import basestring
```

However, there is no `path` file or library, hence returning the following error:
```
ModuleNotFoundError: No module named 'past'
```

Abhayindia:
Fixes #1252.

`dronekit/__init__.py` imports `basestring` from `past.builtins`, but `python-future` (which provides `past`) was never listed as a dependency in `setup.py`. A fresh `pip install dronekit` fails immediately on import.

`basestring` is only needed for Python 2/3 compatibility, so this replaces the import with the same `sys.version_info` gating already used a few lines up for the `collections.abc` change - no new dependency needed.

Confirmed the module is genuinely gone and the import still works:
```
$ python -c "import past"
ModuleNotFoundError: No module named 'past'

$ python -c "import dronekit; print('import OK')"
import OK
```


Abhayindia:
@hamishwillee any chance of a look at this when you have a moment? it's a one-line removal of an undeclared `past.builtins` import that breaks a fresh install on py3.
happy to close it if the repo isn't taking changes right now - just didn't want to leave it sitting silently.

Abhayindia:
tested this properly rather than just eyeballing it: in a clean py3.12 venv with `monotonic` + `pymavlink` and **no** `past` installed, master fails at `dronekit/__init__.py:51` with `ModuleNotFoundError: No module named 'past'`, while this branch imports fine and gives `basestring is str` -> True. `sys` is already imported at line 35, so the version guard is safe, and `basestring` is genuinely used at lines 1649 and 2404.
so this is the whole "fresh install is broken on py3" story in one line. no rush on it, just wanted to show the check rather than assert it.

